PR #2, opened 2019-10-08
#5459: React transpiling

creido-welly:
- added jest and enzyme for component testing
- npmignore added for package configuration


vladzhigulsky (approved):
👍

PR #10, opened 2019-11-11
#5607 feat: add Sass structure

chriddell:
Implements https://github.com/wellcometrust/corporate/issues/5607.

Adds a basic Sass structure to the project.

Features:

- non-component styles housed in separate folder
- mixins/variables/etc are shared between files
- full control over `@imports` (and therefore order of style declarations)
- now using `rollup-plugin-postcss` to build CSS (this is because it watches the full Sass dependency tree for changes, whereas other rollup plugins only watched JS dependency tree/not sub-dependencies e.g. Sass partials)

Lots of scope to expand the `styles` folder, however kept simple for now.


rollup.config.js
```diff
@@ -48,16 +47,10 @@ export default {
...
-      options: {
-        outputStyle: 'compressed'
-      },
-      processor: css => postcss([autoprefixer])
-        .process(css)
-        .then(result => result.css)
-    }),
-
-  ]
+    postcss({
```
creido-welly, line 50: Nice. I've always used postcss to supplement SASS dependencies as opposed to doing the whole compilation job itself. Simplifies things 👍

src/styles/_base.scss
```diff
@@ -0,0 +1,354 @@
+// ----------------------------------
```
creido-welly: As this is a library we need to be careful about how we apply reset rules given that the project consuming the library would have its own set. When we first included the new header from corporate-react in the corporate site we noticed we were duplicating resets and in some cases overriding the existing styles. It would be better to include any reset rules directly in the component classes themselves so we have no side effects.

chriddell: Good point.

I was mostly including resets here as a proof of concept, to ensure styles were imported in the order we wanted and that everything was bundled correctly. I can actually remove this file as the POC has been proven

package.json
```diff
@@ -2,8 +2,8 @@
   "name": "corporate-components",
   "version": "1.0.0",
   "description": "Component library for main Wellcome corporate site",
-  "main": "dist/esm/index.js",
-  "module": "dist/esm/index.js",
+  "main": "dist/index.js",
```
creido-welly: I think this is probably fine as we'll be using the library internally as ES modules.

FYI, the reason for `esm` was in the case we would want to export in **commonjs** AND **es modules**. A properly setup library (for public consumption) prob should cater for both with a package.json configured in similar way to the following:

```
// package.json
...
  "main": "dist/index.js",
  "module": "dist/esm/index.js",
...
```

chriddell: Ah, thanks for flagging. Now I'm looking at it, I'm not sure why I removed the `module` value... 😬

I updated the `main` as the postcss plugin dumps the compiled CSS to wherever the `main` entry is. I could set the `entry` option on the postcss plugin in the rollup config, though, and we could keep the `esm` folder?

creido-welly: Ok cool. Let's keep `esm` to make it a little clearer that that's what's available. We can add tooling to export commonjs too in future if we think it's needed

rollup.config.js
chriddell: Issue I had with other plugins (`rollup-plugin-sass` & `rollup-plugin-scss`) is that they weren't watching the Sass partials for changes (which meant the Sass wouldn't compile on the `rollup -w[atch]` task).

Maybe there's a way around this by adding another plugin to handle watching the (Sass) dependency tree? Not sure what the best option is at the mo.

vladzhigulsky (approved):
Looks good to me.

PR #13, opened 2019-11-14
#5590 feat: add typescript (WIP)

creido-welly:
## What
Following build tools have been updated to accept typescript extensions `.ts` and `.tsx`
- Storybook
- build process
- JS linting
- Test framework (Jest)

**N.B. React component files can have either a `.js` or `.jsx` extension but with TypeScript the file extension has to be specifically named `.tsx`**


.eslintrc.js
```diff
@@ -0,0 +1,72 @@
+module.exports = {
```
creido-welly, line 1: `.eslintrc` without an extension has been deprecated and should either be .json or .js with the latter being preferred

chriddell (approved):

.eslintrc.js
vladzhigulsky: Good one.

vladzhigulsky (approved):

vladzhigulsky (approved):

PR #16, opened 2019-11-19
#5651 feat(refactor): allow absolute imports

chriddell:

Allow absolute imports of common modules (e.g. components, styles) to save us writing relative paths `../../` each time.

## What does it look like?

- Updates `rollup.config.js` with `customResolveOptions` to absolute import locations.
- Updates `.eslintrc.js` with `moduleDirectory` of absolute import locations
- Imports relevant modules absolutely within components

## Have the following been considered/are they needed?
- [ ] Spoken to the right people?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -65,7 +65,8 @@ module.exports = {
   "settings": {
     "import/resolver": {
       "node": {
-        "extensions": [".js", ".jsx", ".ts", ".tsx"]
+        "extensions": [".js", ".jsx", ".ts", ".tsx"],
+        "moduleDirectory": ["node_modules", "./src", "./src/components"]
       }
     }
   },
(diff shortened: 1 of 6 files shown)


creido-welly (commented):
It would be good to know the reasoning behind this i.e. pros and cons. I'm more than happy to adopt it if it makes our lives easier

chriddell:
@creido-welly 

As I see it...

**Pros**
- Will decrease time spent on refactoring relative imports as project grows (e.g. component becoming sub-component, or vice-versa - _Find and replace_ should be easier)
- Improves readability
- Typing '../' repeatedly is slightly annoying 🙃 (and it's not always immediately clear how many levels are required)

**Cons**
- We can't have any custom components with the same name as any external components/libraries (although we can resolve this if we only resolve modules from `src` as opposed to `src/components`

creido-welly (approved):

PR #17, opened 2019-11-19
#5607 feat(refactor): add Sass structure

chriddell:

Spikes a Sass structure as discussed in #10. Useful to start a discussion in to how we should structure our styles.

## What does it look like?

- Uses ITCSS
- JS Component (folders) own the styles; they are imported into a main SCSS file (`app.scss`)
- Other Sass (mixins, settings, etc.) is contained within a separate `styles` folder

---

See also #18


chriddell:
Superseded by https://github.com/wellcometrust/corporate-components/pull/19

PR #18, opened 2019-11-19
#5607 feat(refactor): add Sass structure [option 2]

chriddell:

Spikes a Sass structure as discussed in #10. Useful to start a discussion in to how we should structure our styles.

## What does it look like?

- JS Component (folders) own their own styles
  - `__style.scss` file at component root...
  - ...which is `@imported` into a main SCSS file (`app.scss`)
  - Component styles import relevant "base" styles (mixins/settings/extends etc.) from `styles` folder
- `styles` folder does not write styles, only partials to be `@imported` to components

---

See also #17


chriddell:
Superseded by https://github.com/wellcometrust/corporate-components/pull/17

PR #19, opened 2019-11-20
#5607 feat: implement Sass structure

chriddell:

Implements a more opinionated Sass structure to the project as discussed in https://github.com/wellcometrust/corporate/issues/5607.

## What does it look like?

- Strip styles folder to essentials: tools, settings and components
- Remove extraneous folders
- Move _base.scss to app.scss, comment out unused folders, explain decision - scope to expand structure as seen in @import statements, ITCSS etc.
- Remove @import './src/assets/styles/base'; from component styles
- Import component styles to ./src/assets/styles/30-components/__manifest

## Have the following been considered/are they needed?
- [x] Docs?
- [x] Spoken to the right people?


src/assets/styles/10-settings/_breakpoints.scss
```diff
@@ -0,0 +1,14 @@
+// ----------------------------------
+// UI Settings
+// Breakpoints
+// ----------------------------------
+// Design System 1.0 Alpha
+// 2019-11-11
+// ----------------------------------
+
+$breakpoints: (
```
creido-welly, line 9: 👍 good to see `sm`, `md` etc as opposed to `mobile`, `tablet` etc

src/components/AComp/_acomp.scss
```diff
@@ -9,14 +11,14 @@
   z-index: 1;
 }
 
-.acomp__title {
-  color: white;
-}
+  .acomp__title {
```
creido-welly, line 12: I'm assuming you've added the indents here to indicate descendant selectors of the main component class. I think it maybe unnecessary as the BEM format will communicate that, plus we're aiming to nest the selectors as little as possible so the SASS should perhaps indicate that?

src/assets/styles/app.scss
```diff
@@ -0,0 +1,10 @@
+// ----------------------------------
+// UI
+// ----------------------------------
+// Design System 1.0 Alpha
+// 2019-11-11
+// ----------------------------------
+
```
creido-welly, line 7: Could we insert `@import './src/assets/styles/base';` here instead of inside each of the component files or would that break the structure?

chriddell: I did think about this (albeit briefly), but it's tough to come to a decision. Factors at play:

- what happens if/when the rest of the Sass folders start emitting styles?
- does `@import`ing `_base.scss` in to `app.scss`, along with the component styles confuse the ordering/structure of styles?
- do the above cases create an argument for breaking down `styles/` in to two folders - one which **does** emit CSS, and one which **doesn't** (ie. `styles/base` & `styles/utils`)

I feel like, for now, the current method makes sense as we're not emitting any CSS from the `styles/` folders and `@import`ing `_base.scss` in to individual components makes it explicitly clear that the component relies on the base theme settings.

src/assets/styles/10-settings/_breakpoints.scss
chriddell: Agreed - think it's always best to shorthand when it's widely accepted syntax 👍🏻

src/assets/styles/app.scss
creido-welly: Lots of maybes and things to consider so probably worth us having a more in-depth chat about this to go through the options

src/components/AComp/_acomp.scss
chriddell: > assuming you've added the indents here to indicate descendant selectors

Yep. 

> we're aiming to nest the selectors as little as possible

With both our browser support needs (IE11...) and the messy realities of writing CSS do you think it's likely that we will need to write selectors that nest? If that is the case, I think the indents give us good scope to handle this. If we look at the following example...

```css
.component {}

.component__title {}

.component__body {}

.component__body-title {}
```

With BEM, it looks like `.c-component__body-title` is an element of the `.c-component` block, but in reality it's dependent on the `.c-component__body` block. Whilst this does break an explicit BEM rule, I've found that sometimes it is necessary to do so to work with unwieldy markup or even just to get the desired design. Indenting would help us see immediately the BEM model, even whilst breaking (or _embellishing on_ 😇) a BEM rule.

(I would personally want to write the above in the following way):

```css
.component {}

  .component__title {}

  .component-body {}

    .component-body__title {}
```

---

Sidnote: Even ignoring the above arguments - for me the indents help as they make it immediately clear the hierarchy of selectors, embellishing on the BEM conventions!

Interested to hear your thoughts?

src/assets/styles/10-settings/_breakpoints.scss
creido-welly: It also ties in with the work Vlad did on typefaces and spacings 👍

src/assets/styles/app.scss
chriddell: Happy with that - can include indents from https://github.com/wellcometrust/corporate-components/pull/19#discussion_r349021872 in that chat, too? Though more of a simpler issue of preference so should be more straight-forwards to resolve

chriddell: See https://github.com/wellcometrust/corporate-components/pull/19#issuecomment-557049354 for results of discussion.

chriddell:
As per discussion with @creido-welly...

## To do:

### Structure

- [x] Strip `styles` folder to essentials: `tools`, `settings` and `components`
- [x] Remove extraneous folders
- [x] Move `_base.scss` to `app.scss`, comment out unused folders, explain decision - scope to expand structure as seen in `@import` statements, ITCSS etc.
- [x] Remove `@import './src/assets/styles/base';` from component styles
- [x] Import component styles to `./src/assets/styles/30-components/__manifest`

Also discussed was the scope to expand to a `layouts` style folder, as and when the project requires.

###  Indents

- [ ] Test preferred indent style with stylelint

src/components/AComp/_acomp.scss
chriddell: Resolved to using indents providing they play nicely with stylelint https://github.com/wellcometrust/corporate/issues/5672

creido-welly (approved):

PR #20, opened 2019-11-21
#5662 fix: Icon TypeScript

creido-welly:
- fixed mismatched type error in Icon
- applied type for all SVG Icon components

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "scripts": {
     "build:clean": "rm -rf dist",
     "build": "npm run build:clean && rollup -c ./config/rollup.config.js",
-    "develop": "rollup -c ./config/rollup.config.js -w",
+    "dev": "npm run build -- -w",
     "lint": "npm run lint:js",
     "lint:js": "eslint --fix ./src --ext .js,.jsx,.ts,.tsx && echo 'JS linting complete'",
     "prepare": "npm run build",
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -1,30 +1,33 @@
 import React from 'react';
 
-// import iconTypes from './iconTypes';
-import CircleCross from './components/CircleCross';
+import iconMapping from './iconMapping';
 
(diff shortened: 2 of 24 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
   "scripts": {
     "build:clean": "rm -rf dist",
     "build": "npm run build:clean && rollup -c ./config/rollup.config.js",
-    "develop": "rollup -c ./config/rollup.config.js -w",
+    "dev": "npm run build -- -w",
```
chriddell, line 19: What does the `--` on its own do?

src/components/Icon/Icon.tsx
```diff
@@ -1,30 +1,33 @@
 import React from 'react';
 
-// import iconTypes from './iconTypes';
-import CircleCross from './components/CircleCross';
+import iconMapping from './iconMapping';
 
-type IconProps = {
+type IconType = {
```
chriddell: Any reason this is called `IconType` as opposed to `IconProps`? Is it confusing that `ButtonProps` are still called `ButtonProps` (consistency)?

package.json
creido-welly: it's used to pass in the `-w` param to the underlying script 

so effectively it runs `npm run build:clean && rollup -c ./config/rollup.config.js" -w`

src/components/Icon/Icon.tsx
creido-welly: True we should be consistent. There's no consensus of opinion so I suggest the following

If the type definition is describing component props we use `...Props` suffix and otherwise use `...Type` to describe anything else

chriddell: Sounds good to me - what would the "anything else" be, out of curiosity? Parameters for a function?

creido-welly: Could be, or any variable we need to declare which TS flags up as an issue

chriddell (approved):

PR #24, opened 2019-11-25
Feat/merge typography and spacings #5604

creido-welly:
Merging Vlad's work on responsive styles for typography and spacings


src/assets/styles/10-settings/_breakpoints.scss
```diff
@@ -6,9 +6,23 @@
...
-  md: 738px,
-  lg: 920px,
-  xl: 1100px
+  sm: 48rem,
+  md: 64rem,
+  lg: 75rem,
+  xl: 84rem
 ) !default;
+
+@mixin breakpoint($breakpoint) {
```
chriddell, line 24: Thoughts on using sass-mq going forwards? https://github.com/sass-mq/sass-mq

Gives us a more robust media-queries and we don't have to maintain it ourselves 😄

Happy to raise an issue/PR for this separately though.

src/assets/styles/10-settings/_spacing.scss
```diff
@@ -0,0 +1,48 @@
...
+  // --space-lg: 24px
+  // --space-xl: 48px
+  @include breakpoint(sm) {
+    --space-sm: calc(0.5 * var(--space-unit));
+    --space-md: calc(0.75 * var(--space-unit));
+    --space-lg: calc(1.5 * var(--space-unit));
+    --space-xl: calc(3 * var(--space-unit));
+  }
+
+  // --space-md: 16px
```
chriddell, line 35: Do you think we should keep the comments detailing pixel values? Means we have to update them if/when the sizes change (although I'm not sure they will change much, if at all...)

src/assets/styles/10-settings/_spacing.scss
```diff
@@ -0,0 +1,48 @@
...
+  // --space-lg: 32px
+  // --space-xl: 64px
+  @include breakpoint(md) {
+    --space-md: calc(1 * var(--space-unit));
+    --space-lg: calc(2 * var(--space-unit));
+    --space-xl: calc(4 * var(--space-unit));
+  }
+}
+
+// USAGE
```
chriddell, line 45: Probably a good time to discuss comment formatting... any preferences?

My preferred is:

- Single-line comment `//` for anything ~60 chars and under
```css
.c-component--fixed {
  // when `fixed`, overlay needs to be above header
  z-index: 901;
}
```
- Multi-line block for anything that goes over ~60 characters
```css
/**
 * I am a comment that goes over 60 characters so Chris has
 * studiously created a multi-line block comment. Thanks Chris!
 */
```
- Single-line block for a delimiter
```css
/* Headings */

h1 {
  ...
}

h2 {
  ...
}

h3 {
  ...
}
```

chriddell: Again, happy to approve this PR and deal with this separately if that gets us moving.

chriddell (approved):

src/assets/styles/10-settings/_breakpoints.scss
creido-welly: Absolutely let's use it. Just include it with next style update you do, no need for another ticket

src/assets/styles/10-settings/_spacing.scss
creido-welly: It's good for quick reference, esp if it means not opening Zeplin each time

They should be updated at the same time as the actual values

src/assets/styles/10-settings/_spacing.scss
creido-welly: Sure, let's talk about it offline

creido-welly: Let's chat about it offline

PR #39, opened 2020-01-06
Feat/add navlinks #5848

creido-welly:


src/components/Nav/_nav-item.scss
```diff
@@ -16,7 +16,7 @@
   }
```
chriddell, line 18: Lots of nesting going on in this file... any room for refactoring the classnames to use modifiers? Can add to backlog if we want to get this merged asap.

src/components/Nav/_nav-item.scss
```diff
@@ -112,6 +83,35 @@
...
+    padding: calc(2 * var(--space-unit));
+
+    &:active,
+    &:focus,
+    &:hover,
+    &.active {
+      color: var(--colour-white);
+    }
+
+    .nav__link-text:before {
```
chriddell, line 104: personally would write these with media queries inside the nested selector, to avoid repeated selectors, e.g.

```css
.top-level-selector {
  color: blue;

  @media (min-width: 12px) {
    color: black;
  }
 
  .nested-selector {
    color: blue;
    
    @media (min-width: 12px) {
      color: black;
    }

    @media (min-width: 24px) {
      color: red;
    }
  }
}
```

src/components/Nav/_nav-item.scss
creido-welly: yeah it's a nasty file. I've added a TODO at the top so will create a backlog ticket

chriddell (approved):

PR #42, opened 2020-01-09
Feat/add header wrapper #5725 (WIP)

creido-welly:
Header and associated files have been moved across and all is working aside from the sticky header mechanism. Additional refactoring will follow to ensure better class naming and sass structure.


src/assets/styles/10-settings/_global.scss
```diff
@@ -8,3 +8,7 @@
 
 $text-base-pixels: 16;
 $max-width: map-get($breakpoints, lg);
+
+.visually-hidden {
```
chriddell: I think this is in the wrong place. We should add a "utils" folder to the Sass structure imo (e.g. `40-utilities/`), otherwise this file could get very messy. 

(sidenote: I've actually done this already in a feature branch I'm working on, as I had a need for `visually-hidden` too!)

chriddell: I also think we should prefix the utility classes (`u-`)

src/components/Header/Header.md
```diff
@@ -0,0 +1,22 @@
+# Header
```
chriddell, line 1: 👍🏻👍🏻👍🏻

chriddell (changes requested):
Seems like styles are getting a bit unwieldy already... I'll take a look at how we can reduce the nesting and hopefully better structure placeholder classes.

Also I've added sass-mq in a feature branch I'm working on, and that can handle max-width media queries too.

creido-welly:
> Seems like styles are getting a bit unwieldy already... I'll take a look at how we can reduce the nesting and hopefully better structure placeholder classes.
> 
> Also I've added sass-mq in a feature branch I'm working on, and that can handle max-width media queries too.

I hear you but like I said the style work is WIP and this is **not** how I intend to leave it. This is the first stage in firstly getting the components working in this repo and then refactoring in subsequent PR(s).

src/assets/styles/10-settings/_global.scss
creido-welly: Done

chriddell (approved):

PR #44, opened 2020-01-09
Feat/add searchform #5724

chriddell:
Implements https://github.com/wellcometrust/corporate/issues/5724.

Readme (`SearchPane.md`) to follow in separate PR.


chriddell:
~I appreciate there are files which have been touched in this PR which ideally should have been dealt with separately (all style-related, particularly sass-mq and the button styles).~

~In future will chat about these things prior to including in PR, to see if we should create a separate PR 🙃~

Ignore above, going to amend this PR to deal with styles separately. @creido-welly has done a lot of it in https://github.com/wellcometrust/corporate-components/pull/43 so should be straightforwards.

src/components/SearchPane/SearchPane.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+  SearchPaneContextProvider
+} from './Context/SearchPaneContext';
+
+const SearchPaneExample = () => (
+  <SearchPaneContextProvider>
+    <SearchPaneContext.Consumer>
+      {({ isSearchActive, toggleSearch }) => (
+        <>
+          {!isSearchActive && (
+            <button onClick={() => toggleSearch(true)} type="button">
```
creido-welly: Picky nicey nicey:  could use the standard Button component here :)

src/components/SearchPane/Controls/SearchPaneControls.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  const closeAll = () => {
+    toggleNav(false);
+    toggleSearch(false);
+  };
+
+  return (
+    <div className="search-pane__controls">
+      <Button
+        className="search-pane__close"
```
creido-welly: `search-pane__btn-close` might be a better descriptor

src/components/SearchPane/Controls/SearchPaneControls.tsx
```diff
@@ -0,0 +1,40 @@
...
+        type="button"
+        onClick={closeAll}
+        icon="close"
+        styled={false}
+      >
+        Close <span className="u-visually-hidden">search</span>
+      </Button>
+      <Button
+        className="search-pane__back"
```
creido-welly: `search-pane__btn-back` might be a better descriptor

src/components/SearchPane/Controls/SearchPaneControls.tsx
```diff
@@ -0,0 +1,40 @@
...
+        type="button"
```
creido-welly: Prob worth setting `type = button` as default in Button component itself, it would save us having to set this prop each time the component is used

src/components/SearchPane/_search-pane.scss
```diff
@@ -0,0 +1,210 @@
...
+// Design System 1.0 Alpha
+// 2020-01-20
+// ----------------------------------
+
+// Variables
+
+$transition-duration: 0.4s;
+
+// Extends
+
```
creido-welly: Looks like we have some duplication with `%nav-btn-base` in `_header.scss`

Doesn't need addressing in this PR but we should discuss amalgamating the two

src/components/Button/Button.tsx
```diff
@@ -24,11 +25,15 @@ export const Button = ({
...
   tabIndex,
-  type
+  type,
+  styled = true
 }: ButtonProps) => {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const Element: any = href ? 'a' : 'button';
-  const classNames = cx('btn', { [`${className}`]: className });
+  const classNames = cx({
+    btn: styled,
```
creido-welly, line 34: 👍

src/components/SearchPane/Controls/SearchPaneControls.tsx
chriddell: Good point, will do.

src/components/SearchPane/_search-pane.scss
```diff
@@ -0,0 +1,210 @@
...
+    padding: 0;
+  }
+
+  @include breakpoint(md) {
+    color: var(--colour-white);
+    margin-right: var(--space-lg);
+  }
+}
+
+// .search-pane__form-submit .btn__text {
```
creido-welly: remove if no longer needed

creido-welly (approved):
Good work, really clean and easy to read. Just a couple of minor points to look at and all good to merge

src/components/SearchPane/_search-pane.scss
chriddell: is needed - however I'm going to remove and add in separate PR (as then will be able to use visually-hidden mixin/extend 🚀)

src/components/Button/Button.tsx
```diff
@@ -41,7 +46,7 @@ export const Button = ({
         }
       }}
       tabIndex={tabIndex}
-      type={!href ? type : null}
+      type={!href ? type : 'button'}
```
creido-welly: This needs to be left as `type={!href ? type : null}`

If an href is specified the component is rendered as an <a> as opposed to a <button>

modify line 28 instead to be `type = 'button',`

chriddell: 🤦🏻‍♂️ of course! I've misread the conditional - will update

chriddell: fixed

creido-welly (approved):

PR #45, opened 2020-01-09
#5463 - feat: search result list component.

DipakShrestha18:


chriddell (changes requested):
Think `ResultList.stories.txt` has the wrong extension - should it be `.tsx`?

DipakShrestha18:
@chriddell changed the extension.

chriddell (approved):

PR #48, opened 2020-01-10
#5463 - feat: adding ResultItem story, styling and some refactoring.

DipakShrestha18:


src/components/ResultItem/ResultItem.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+type ResultItemProps = {
+  className?: string;
+  meta?: string;
+  title: string;
+  href: string;
+  text?: string;
```
chriddell: Can we do these in alphabetical order please? Just to stay consistent (realised I need to update some components I've written to do this, too!)

Only exception would be a `props` prop, which would handle any "other" props (see the Hamburger component for an example)

chriddell (changes requested):
I'm also gonna pull this down and fix the rem() mixin before we merge

src/components/ResultItem/ResultItem.tsx
DipakShrestha18: Done.

src/components/ResultItem/_resultItem.scss
```diff
@@ -0,0 +1,84 @@
+@mixin font-size-with-line-height($fontSize, $lineHeightPx) {
```
chriddell: filename should be kebab case (`result-item.scss`)

chriddell (changes requested):

src/components/ResultItem/_resultItem.scss
DipakShrestha18: done.

src/assets/styles/30-components/__manifest.scss
```diff
@@ -13,3 +13,5 @@
 @import './src/components/Nav/nav-list';
 @import './src/components/SearchPane/search-pane';
 @import './src/components/WellcomeCollectionBanner/wellcome-collection-banner';
+@import './src/Components/Logo/Logo';
+@import './src/components/ResultItem/ResultItem';
```
chriddell: path needs updating due to filename change

chriddell (changes requested):

src/assets/styles/30-components/__manifest.scss
DipakShrestha18: done.

chriddell (commented):
All looks good except styles - I'm gonna work on them as would be great to have done as part of this PR - so likely have to wait til Monday to merge this.

DipakShrestha18:
No probs.

chriddell (approved):

PR #49, opened 2020-01-10
Feat/add global style helpers #5886

chriddell:
Implements https://github.com/wellcometrust/corporate/issues/5886


src/assets/styles/20-tools/__manifest.scss
```diff
@@ -2,9 +2,9 @@
 // UI Tools
 // ----------------------------------
 // Design System 1.0 Alpha
-// 2019-11-11
+// 2019-01-10
```
DipakShrestha18: 2020??

DipakShrestha18 (approved):

src/assets/styles/20-tools/__manifest.scss
creido-welly: Design system date is based on Jason's design work so we should leave the data as it was

chriddell: Ah okay, I'll update back to 2019-11-11 everywhere 👍🏻

creido-welly (approved):

PR #55, opened 2020-01-13
#5462 - feat: Search result page header added.

DipakShrestha18:


src/components/index.ts
```diff
@@ -13,6 +13,7 @@ export { WellcomeCollectionBanner } from 'WellcomeCollectionBanner/WellcomeColle
 export { Logo } from 'Logo/Logo';
 export { ResultItem } from 'ResultItem/ResultItem';
 export { SearchPane } from 'SearchPane/SearchPane';
```
creido-welly, line 16: please order alphabetically

src/components/PageHeader/_page-header.scss
```diff
@@ -0,0 +1,99 @@
...
+  @include mq(sm) {
+    padding-bottom: rem(32);
+  }
+
+  @include mq(md) {
+    padding-bottom: rem(64);
+  }
+}
+
+// DEPRECATED
```
creido-welly: If we get clear deprecation flags like this it's good to just remove as they'll be very unlikely to make it into the new codebase (if they do there needs to be a very good reason)

src/components/index.ts
DipakShrestha18: done.

src/components/PageHeader/_page-header.scss
DipakShrestha18: done.

src/components/PageHeader/PageHeader.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+export const PageHeader = ({
+  title,
+  className,
+  standfirst,
+  meta
+}: PageHeaderProps) => {
+  return (
+    <div className="title-banner" {...className}>
```
creido-welly: Adding className like this won't work. Take a look at `src/components/Hamburger/Hamburger` for an example using the **classnames** utility

i.e.
```
const cssClasses = classnames('btn--nav-toggle', {
    [`${className}`]: className
  });

...
<div className={cssClasses}>
```

This will ensure the default classNames are applied along with anything extra passed thru as props

DipakShrestha18: Thanks, I wasn't sure about this one. It's good that you picked up. 

This is done too.

creido-welly (approved):

PR #57, opened 2020-01-13
Feat/add section component #5880

chriddell:
Implements https://github.com/wellcometrust/corporate/issues/5880.


src/components/Section/Section.md
```diff
@@ -0,0 +1,19 @@
...
+
+## When to use
+
+A Section can be used multiple times within a page/layout.
+
+## How to use
+
+A Section must contain child components, which the section will wrap in either `<div>` or `<section>`, depending on the `type` prop.
+
+## Accessibility notes
```
creido-welly: I'm guessing this was copied from Header.md?

chriddell: Ah, thanks for flagging - fixed in new commit.

src/assets/styles/20-tools/_extends/_colours.scss
```diff
@@ -0,0 +1,94 @@
...
+// Extends
+// Colours
+// ----------------------------------
+// Design System 1.0 Alpha
+// 2019-11-11
+// ----------------------------------
+
+// Background Colours
+
+%bg-colour-black {
```
creido-welly: Not sure how useful these extends are. If there were more style rules to be applied with a more abstracted name it would be worth it certainly

src/components/Section/Section.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type SectionProps = {
+  children: React.ReactNode;
+  bgColor?: boolean;
+  className?: string;
+  type?: 'div' | 'section';
+};
+
+export const Section = ({
+  bgColor,
```
creido-welly: re naming and application of different styles (bgColor and section--has-bg-color), I think we need to abstract this more as it could be about more than simply background colour

current site uses `section--odd` which is part the way there but isn't necessarily odd either :)

`section--alternate` would be a better fit for that

I can see bgColor is boolean but its name suggests differently, so perhaps `hasAlternateStyle`?

src/assets/styles/20-tools/_extends/_colours.scss
chriddell: Yeah that's fair enough, they would be useful as utility classes I think but I also think it's best not to have too many utility classes in pattern library.

Happy enough parking this for now 👍

src/components/Section/Section.tsx
chriddell: Yep got this - it is solely about background color at the mo though isn't it?

Style from corporate:

```css
.section--odd {
  background-color: $blue05;
}
```

Maybe best to not over-engineer it if it's only setting a bg color for the time being - happy enough to rename the prop and/or class though, but think it should be descriptive (on both counts) `hasBgColor` & `.section--has-bg-color`?

src/assets/styles/20-tools/_extends/_colours.scss
chriddell: Actually now I come to think of it - this would save us lines of code in the built CSS file, so could be worth doing after all from a performance POV?

src/components/Section/Section.tsx
creido-welly: It is solely background colour at the moment but it's also a little too close to the old html attribute bgcolor and <b> and <i> elements

creido-welly (approved):

PR #59, opened 2020-01-14
Feat/search pane vertical content shift #5724

creido-welly:


src/components/SearchPane/SearchPane.tsx
```diff
@@ -9,10 +9,37 @@ import SearchContext from 'SearchPane/Context/SearchPaneContext';
 import SearchPaneForm from './Form/SearchPaneForm';
 import SearchPaneControls from './Controls/SearchPaneControls';
 
+const CSS_CLASSES = {
+  IS_SEARCH_ACTIVE: 'search-open'
```
chriddell: A bit pernickety from me - but what do you think about renaming the selector to something stateful - e.g. `has-search-open`, `has-open-search-pane`, `has-open-search`?

creido-welly: I'll change to `is-search-active` (we're using `...active` for other selectors)

creido-welly: Done

chriddell (approved):

config/storybook/styles/_storybook-global.scss
```diff
@@ -43,3 +43,17 @@ h5 {
...
+
+// stylelint-disable selector-max-id
+.is-search-active .header,
+.is-search-active .image-lead,
+.is-search-active .pride-strip,
+.is-search-active .banner,
+.is-search-active .main,
+.is-search-active #main,
+.is-search-active .footer {
+  // TODO: figure out why we can't use @include mq(md) here
```
chriddell: Caused by the order of imports in storybook-app.scss.

The below will fix it:

```css
@import 'assets/styles/app';
@import './storybook-global';
```

chriddell (changes requested):
See comment re: mq() in storybook.

config/storybook/styles/_storybook-global.scss
creido-welly: Sorted, worked fine with all of the extends ?!?!

chriddell: Doesn't matter with extends it seems, they can be defined anywhere. I didn't know that, but good to find out!

chriddell (approved):

PR #62, opened 2020-01-15
#5920 feat: adds Grid storybook component

chriddell:
Implements https://github.com/wellcometrust/corporate/issues/5920

### DIFF
--- a/config/storybook/styles/_storybook-global.scss
+++ b/config/storybook/styles/_storybook-global.scss
@@ -56,3 +56,13 @@ h6 {
     transform: translate3d(0, 64px, 0);
   }
 }
+
+// Grid
+.st-grid-cell-visualiser {
+  align-items: center;
+  background-color: var(--colour-concrete-20);
+  border: 1px solid;
+  display: flex;
+  height: 200px;
+  justify-content: center;
+}
--- a/src/components/Grid/Grid.stories.tsx
+++ b/src/components/Grid/Grid.stories.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { select } from '@storybook/addon-knobs';
+
+import Grid, { GridCell } from './Grid';
+
+const GridExample = () => {
+  const columnCount: 1 | 2 | 3 = select('column count', [1, 2, 3], 3);
+
+  /**
+   * Creates an array of GridCell components based
+   * on the given columnCount from the Storybook select knob.
+   *
+   * @return {array} - array of React Components
+   */
+  const createGridCells = () => {
+    const cells = [];
+
+    for (let i: 1 = 1; i <= columnCount; i += 1) {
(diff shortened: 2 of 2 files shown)


config/storybook/styles/_storybook-global.scss
```diff
@@ -56,3 +56,13 @@ h6 {
     transform: translate3d(0, 64px, 0);
   }
 }
+
+// Grid
+.st-grid-cell-visualiser {
```
creido-welly, line 61: 👍

chriddell: Likely we'll have to split this file in to more partials as Storybook grows, `st-` seemed like a good starting point for namespacing 🤘🏻

src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const columnCount: 1 | 2 | 3 = select('column count', [1, 2, 3], 3);
+
+  /**
+   * Creates an array of GridCell components based
+   * on the given columnCount from the Storybook select knob.
+   *
+   * @return {array} - array of React Components
+   */
+  const createGridCells = () => {
+    const cells = [];
```
creido-welly, line 17: Juicy titbit for yer, if you wanna "func programming" it up you could also do

```
const cells = Array.from(Array(columnCount), (x, index) => {
      const i = index + 1;

      return (
        <GridCell column={i} columnCount={columnCount}>
          <div className="st-grid-cell-visualiser">
            {`Grid Cell ${i} of ${columnCount}`}
          </div>
        </GridCell>
      );
    });
```

😬
drop 🎤

creido-welly (approved):

config/storybook/styles/_storybook-global.scss
creido-welly: perhaps, all good for now tho

src/components/Grid/Grid.stories.tsx
chriddell: What do you think about `const i` not being type-checked? It throws this error:

```bash
Type 'number' is not assignable to type '1 | 2 | 3'
```

And trying to set explicit type on `i` (of `1 | 2 | 3`), e.g:

```javascript
const i: 1 | 2 | 3 = index + 1;
```

throws:

```bash
Type 'number' is not assignable to type '1 | 2 | 3'
```

Can't for the life of me figure out how to type-check it properly 😶

creido-welly: have you tried `const i: number = index + 1;` ?

chriddell: (realise we spoke about this IRL but just for others' clarity...) 

This doesn't work as the `column` property on `GridCell` is type-checked as `1 | 2 | 3`, so `i` must explicitly match that. 

I'll work on it a little more later.

PR #74, opened 2020-01-17
#5942 feat(refactor): replace rem() functions with CSS variables

chriddell:
Implements https://github.com/wellcometrust/corporate/issues/5942

### DIFF
--- a/src/components/PageHeader/_page-header.scss
+++ b/src/components/PageHeader/_page-header.scss
@@ -1,18 +1,10 @@
 .title-banner__container {
-  // padding: rem(var(--text-base-size-px) * 2) 0;
-  padding: rem(32) 0;
+  padding-left: var(--space-xl);
+  padding-right: var(--space-xl);
   text-align: center;
 
-  @include mq(sm) {
-    padding: rem(48) 0;
-  }
-
-  @include mq(md) {
-    padding: rem(64) 0;
-  }
-
   .horizontal-columns__center & {
-    padding-top: rem(16);
+    padding-top: calc(2 * var(--space-unit));
   }
 }
 
@@ -28,15 +20,7 @@
   }
 
   .title-banner__standfirst {
-    padding-top: rem(16);
-
-    @include mq(sm) {
-      padding-top: rem(24);
-    }
-
-    @include mq(md) {
-      padding-top: rem(32);
-    }
+    padding-top: var(--space-lg);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/components/PageHeader/_page-header.scss
```diff
@@ -1,18 +1,10 @@
 .title-banner__container {
-  // padding: rem(var(--text-base-size-px) * 2) 0;
-  padding: rem(32) 0;
+  padding-left: var(--space-xl);
```
creido-welly: --space-xl will change size depending on the breakpoint (based on spacings in Wellcome Design System - https://zpl.io/bJWJNgW)

is that what you need?

if these values are static or differ in any way then use multiples of --space-unit

chriddell: Yep, --space-xl is intended.

Here's the settings for it:

```css
:root {
  // 32px
  --space-xl: calc(4 * var(--space-unit));

  @include mq(sm) {
    // 48px
    --space-xl: calc(6 * var(--space-unit));
  }

  @include mq(md) {
    // 64px
    --space-xl: calc(8 * var(--space-unit));
  }
}

Comment (creido-welly) [reply src/components/PageHeader/_page-header.scss]: 👍 great, just checking

File: src/components/PageHeader/_page-header.scss [hunk 2]
```diff
@@ -28,34 +20,22 @@
...
+  padding-bottom: var(--space-xl);
   text-align: center;
 
-  @include mq(sm) {
-    padding-bottom: rem(32);
-  }
-
-  @include mq(md) {
-    padding-bottom: rem(64);
+  @include mq($from: sm, $until: md) {
```
creido-welly: I'm guessing this is an override for --space-xl. Title banner is one of the more recent components which would almost certainly use the new spacings so I'd remove this and let the variable height work as normal (unless it looks wrong?)

chriddell: Sounds good, I'll do a quick visual check to make sure

src/components/ResultItem/_result-item.scss
```diff
@@ -1,13 +1,13 @@
 .result-item {
   border-top: 1px solid var(--colour-concrete-10);
-  padding-bottom: rem(25);
-  padding-top: rem(25);
+  padding-bottom: calc(3.125 * var(--space-unit));
```
creido-welly: All good here but might be worth getting Jason's thoughts on the result item spacings. If we can introduce the new spacing standard quickly and easily then it'll bring things into step with his latest work.

chriddell: Agreed - will do

creido-welly (approved):
Code changes approved - refer older spacings of search result list items to Jason for comment

PR #77, opened 2020-01-17
#5949 - feat: extra meta information added.

DipakShrestha18:


src/components/ResultItem/ResultItem.tsx
```diff
@@ -1,17 +1,32 @@
 import React from 'react';
 
+type meta = {
+  type?: string;
+  date?: Date;
+  lastUpdated?: Date;
```
chriddell: I think we should make these strings for now, although I'm not actually sure what data we will pass to these going forwards.

src/components/ResultItem/ResultItem.tsx
```diff
@@ -1,17 +1,32 @@
...
+  href,
+  text
+}: ResultItemProps) => (
   <div className="result-item">
     <div className="result-item__meta">
-      <span className="result-item__meta-type">{meta}</span>
+      <span className="result-item__meta-type">{type}</span>
+      <span className="result-item__meta-type--date">{date}</span>
+      <span className="result-item__meta-type--last-updated">
+        Update {lastUpdated}
```
chriddell: *Updated

src/components/ResultItem/ResultItem.tsx
```diff
@@ -1,17 +1,32 @@
...
+export const ResultItem = ({
+  meta: { type, date, lastUpdated },
+  title,
+  href,
+  text
+}: ResultItemProps) => (
   <div className="result-item">
     <div className="result-item__meta">
-      <span className="result-item__meta-type">{meta}</span>
+      <span className="result-item__meta-type">{type}</span>
```
chriddell, line 25: className should be `result-item__meta--type` (double hyphen)

src/components/ResultItem/ResultItem.tsx
```diff
@@ -1,17 +1,32 @@
...
+  type?: string;
+  date?: Date;
+  lastUpdated?: Date;
+};
+
 type ResultItemProps = {
   className?: string;
   href: string;
-  meta?: string;
+  meta?: meta;
```
chriddell, line 12: 👍

src/components/ResultItem/ResultItem.stories.tsx
```diff
@@ -5,15 +5,19 @@ import { text } from '@storybook/addon-knobs';
 import ResultItem from './ResultItem';
 
 const ResultListExample = () => {
-  const meta = text('meta text', 'Funding scheme');
+  const meta = {
+    type: text('meta type', 'Funding scheme'),
+    date: text('meta date', '8 July 2016'),
+    lastUpdated: text('meta last updated', '1 November 2018')
+  };
```
chriddell: Does this work? (`npm run storybook`)

src/components/ResultItem/ResultItem.stories.tsx
```diff
@@ -5,15 +5,19 @@ import { text } from '@storybook/addon-knobs';
...
   const title = text('title', 'Senior Research Fellowships');
   const href = text('href', '');
   const content = text(
     'desc',
     'Supporting independent researchers who are emerging as global leaders in their field and want to tackle the most important questions in science.'
   );
 
-  return <ResultItem meta={meta} title={title} href={href} text={content} />;
+  return <ResultItem title={title} href={href} text={content} />;
```
chriddell: We should pass the `meta` props to this component.

src/components/ResultItem/ResultItem.tsx
chriddell: These props will be accessed with `meta.type`, `meta.date` and `meta.lastUpdated` (I think)

chriddell (changes requested):

src/components/ResultItem/ResultItem.tsx
```diff
@@ -1,17 +1,43 @@
...
   <div className="result-item">
     <div className="result-item__meta">
-      <span className="result-item__meta-type">{meta}</span>
+      <span className="result-item__meta-type">{type}</span>
+      <span className="result-item__meta-type--date">
+        {new Intl.DateTimeFormat('en-GB', {
+          year: 'numeric',
+          month: 'long',
+          day: '2-digit'
+        }).format(date)}
```
chriddell: Nice - though I imagine we will probably pass the date prop through as a string for now. (depends on what Solr sends back)

src/components/ResultItem/ResultItem.tsx
```diff
@@ -1,17 +1,43 @@
...
+      </span>
+      <span className="result-item__meta-type--last-updated">
+        Update
+        {new Intl.DateTimeFormat('en-GB', {
+          year: 'numeric',
+          month: 'long',
+          day: '2-digit'
+        }).format(lastUpdated)}
```
chriddell: Same as above - helpful if we pass a date through, but not if it's a string ¯\_(ツ)_/¯

src/components/ResultItem/ResultItem.tsx
```diff
@@ -1,17 +1,43 @@
...
+  meta: { type, date, lastUpdated },
+  title,
+  href,
+  text
+}: ResultItemProps) => (
   <div className="result-item">
     <div className="result-item__meta">
-      <span className="result-item__meta-type">{meta}</span>
+      <span className="result-item__meta-type">{type}</span>
+      <span className="result-item__meta-type--date">
```
creido-welly: rename this to `result-item__meta-date` since it's not a variant of the type e.g. 'Report' but an element of its own

src/components/ResultItem/ResultItem.tsx
DipakShrestha18: I don't think so as it's giving me error if I add meta.

src/components/ResultItem/ResultItem.tsx
creido-welly: You can abstract this and following instance into separate IntlDate component (or similar name).

I presume Solr returns a string (simpler format) which could accept string or date.

You could always accept both formats and check if it's typeof string then convert to a Date object before converting into Intl.Date

src/components/ResultItem/ResultItem.tsx
creido-welly: not necessarily

if it's a modifier of result-item__meta then yes use double hyphen

if it's related to result-item__meta as it is then ...meta-type is appropriate

result-item__meta__type is just daft 🙄 (effectively it couples the markup and the style - less flexible)

DipakShrestha18:
@chriddell @creido-welly ready for re-review.

creido-welly (approved):

PR #78, opened 2020-01-20
#5956 junk(refactor): refactors SearchQueryFetch component

chriddell:
Junk branch which refactors the SearchQueryFetch component in to something simpler (function component).

It's a pretty standard search component really - keeps it's own state (errors, busy, results, query), and runs a POST request with the fetch API.

I think (though not 100% sure) that we need to refactor this a bit to make it re-usable? ie. turn it into a HOC component so that it could:

- use any form (not just SearchForm) as an input
- pass it's state down to any { children } props (allowing the SolrConnector to have *any* children - not specifically SearchForm)

Thoughts @creido-welly? Would suggest pulling this branch down and having a tinker around.


creido-welly:
All good code wise

My only comment is around naming of the **SolrConnector** component. Probably better as **SearchConnector** since the code is already very non-search engine specific and given that we're aiming to decouple services backend from frontend for easy swap out (if we ever wanted to).

creido-welly:
Have you tried running this in corporate-react at all?

Just wondering if there would be any issues with server-side trying to use a client-side API (fetch)

If you do have issues take a look at 'isomorphic-unfetch'

chriddell:
> naming of the **SolrConnector** component. Probably better as **SearchConnector** since the code is already very non-search engine specific and given that we're aiming to decouple services backend from frontend for easy swap out (if we ever wanted to).

Very true - will update as per above.

> If you do have issues take a look at 'isomorphic-unfetch'

Ah, used this before. Haven't used in corp-react yet, and I imagine there _will_ be issues with fetch and SSR, should be fine if it uses unfetch.

chriddell:
Closing because https://github.com/wellcometrust/corporate/issues/5973

PR #79, opened 2020-01-20
#5465 feat: Search Results outline component and pagination (WIP)

creido-welly:
To complete:
- Move into the same folder as ResultItem
- Integrate with search API fetching service (SearchConnector)

### DIFF
--- a/src/components/ResultItem/ResultItem.tsx
+++ b/src/components/ResultItem/ResultItem.tsx
@@ -6,13 +6,13 @@ type meta = {
   lastUpdated?: string;
 };
 
-type ResultItemProps = {
+export interface ResultItemProps {
   className?: string;
   href: string;
   meta?: meta;
   text?: string;
   title: string;
-};
+}
 
 export const ResultItem = ({
   meta: { type, date, lastUpdated },
@@ -28,11 +28,11 @@ export const ResultItem = ({
         Updated {lastUpdated}
       </span>
     </div>
-    <h4 className="result-item__title">
+    <h3 className="result-item__title">
       <a href={href} className="result-item__link">
         {title}
       </a>
-    </h4>
+    </h3>
     <p className="result-item__text">{text}</p>
   </div>
 );
--- a/src/components/ResultItem/ResultItem.tsx
+++ b/src/components/ResultItem/ResultItem.tsx
@@ -6,13 +6,13 @@ type meta = {
   lastUpdated?: string;
 };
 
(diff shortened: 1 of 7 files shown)


DipakShrestha18 (approved):

src/components/ResultItem/ResultItem.tsx
```diff
@@ -6,13 +6,13 @@ type meta = {
   lastUpdated?: string;
 };
 
-type ResultItemProps = {
+export interface ResultItemProps {
```
chriddell: Why `interface` and not `type`?

src/components/ResultsList/ResultsList.stories.tsx
```diff
@@ -0,0 +1,49 @@
+// @ts-nocheck
```
chriddell, line 1: Thoughts on renaming this to `ResultList` (removing the -s, to be aligned with `ResultItem` (or vice-versa - renaming `ResultItem` to `ResultsItem`)?

chriddell: I think maybe we could (in future) structure these two components in the same folder anyway...

src/components/ResultItem/ResultItem.tsx
DipakShrestha18: What's the main difference between interface and type? I read few blogs and most of them uses interface and we are using type but not fully understand the differences.

src/components/ResultsList/ResultsList.stories.tsx
creido-welly: The intention is to combine in a future PR. I'll go with ResultsList and ResultsItem since we talk about "show the **results**"

chriddell (approved):

src/components/ResultItem/ResultItem.tsx
creido-welly: Re the difference this SO answer can tell you so much more about it
https://stackoverflow.com/questions/37233735/typescript-interfaces-vs-types#answer-52682220

I resolved to use type throughout as opposed to interface mainly because of point **5. Declaration merging** . If we allow multiple declarations it would make it much easier to accidentally break something.

In this instance I mistakenly thought I had to use interface to use a declaration in other files but that was incorrect (export is used to do that) so I've reverted to `export type ResultItemProps`

chriddell: Thanks for the link, really helpful

PR #95, opened 2020-01-24
#6004: feat(styles): adds main footer and footer-nav styles

chriddell:
Resolves https://github.com/wellcometrust/corporate/issues/6004

- adds `.footer` "top-level" styles
- adds `.footer-nav` styles

\* **doesn't** add `.footer-social` and `.footer-links` styles (see https://github.com/wellcometrust/corporate/issues/6008 and https://github.com/wellcometrust/corporate/issues/6007)

### DIFF
--- a/src/components/NewsletterForm/_newsletter-form.scss
+++ b/src/components/NewsletterForm/_newsletter-form.scss
@@ -39,7 +39,7 @@
 }
 
 .newsletter-form__item--footer {
-  font-size: var(--text-xs);
+  font-size: 12px;
   order: 4;
 }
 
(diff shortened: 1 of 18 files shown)


src/components/NewsletterForm/_newsletter-form.scss
```diff
@@ -39,7 +39,7 @@
 }
 
 .newsletter-form__item--footer {
-  font-size: var(--text-xs);
+  font-size: 12px;
```
DipakShrestha18, line 42: should this be a variable?

DipakShrestha18 (approved):

src/components/NewsletterForm/_newsletter-form.scss
chriddell: ideally yeah - though we don't have a 12px font-size as a typography variable (see `10-settings/_typography.scss`)...

I tried this using the `--text-xs` variable, but it looks a lot different to the current Footer (and we need to be consistent with the footer styles as we'll be serving one from D7 and this new one!)

PR #101, opened 2020-01-28
#6014 feat: adds submission handlers to newsletter form

chriddell:
Resolves https://github.com/wellcometrust/corporate/issues/6014

- writes fetch request for newsletter form

### DIFF
--- a/src/components/NewsletterForm/NewsletterForm.tsx
+++ b/src/components/NewsletterForm/NewsletterForm.tsx
@@ -105,42 +105,40 @@ export const NewsletterForm = ({ className }: NewsletterFormProps) => {
     [`${className}`]: className
   });
 
-  // Show the success response if the submission has been sent
-  if (responseSuccess) {
-    return (
+  return (
+    // Show the success response if the submission has been sent
+    responseSuccess ? (
       <p className="newsletter-form__response-msg newsletter-form__response-msg--success">
         Thank you. If this is the first time you have subscribed to a newsletter
         from Wellcome, you will receive an email asking you to confirm your
         subscription.
       </p>
-    );
-  }
-
-  return (
-    <form
-      action="https://r1.dmtrk.net/signup.ashx"
-      className={classNames}
-      method="POST"
-      onSubmit={handleSubmit}
-    >
-      <NewsletterFormEmail
-        handleBlur={event => handleEmailBlur(event.currentTarget)}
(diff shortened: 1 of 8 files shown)


src/components/NewsletterForm/NewsletterForm.tsx
```diff
@@ -74,22 +105,24 @@ export const NewsletterForm = ({ className }: NewsletterFormProps) => {
     [`${className}`]: className
   });
 
+  // Show the success response if the submission has been sent
+  if (responseSuccess) {
```
creido-welly: Minor: Perfectly valid to render this way however since this is React the ternary approach might be more consistent

i.e. 
```
return (
  {responseSuccess
    ?
    <response/>
    :
    <form />
  }
)

File: src/components/NewsletterForm/NewsletterFormSubmit.tsx [hunk 2]
```diff
@@ -20,7 +22,7 @@ export const NewsletterFormSubmit = ({
       onClick={handleClick}
       type="submit"
     >
-      Subscribe
+      {busy ? 'Submitting' : 'Subscribe'}
```
creido-welly, line 25: Nice

src/components/NewsletterForm/NewsletterForm.tsx
chriddell: Fair point - does this mean we aim to exclusively use ternary operators, then?

creido-welly: Certainly for react component render (since that's what the React docs state should be used) but I don't think we should aim for it exclusively. In instances where code is made less readable then an `if` or `if else` or even refactoring would probably be better

creido-welly (approved):

src/components/NewsletterForm/NewsletterForm.tsx
chriddell: had a brief look at the react docs after you posted, but all I could find was:

> Just like in JavaScript, it is up to you to choose an appropriate style based on what you and your team consider more readable.
> - https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator (beneath code snippets)

tbh I'm happy with whatever is most readable too so it's not really an issue, more just curious as to if React do say "use this way" somewhere!

creido-welly: Fair enough. I couldn't find the reference either :D

PR #104, opened 2020-01-28
#6019 feat: Formatted date component added

creido-welly:


src/components/DateTime/DateTime.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { format, isValid } from 'date-fns';
+
+type DateTimeProps = {
+  dateString: string;
+};
+
+export const DateTime = ({ dateString }: DateTimeProps) => {
```
chriddell: V nice - though it's not (currently) returning anything with _time_ specifically so would it be more appropriately named just `Date` (or something like `FormattedDate`)?

creido-welly: Fair point. I wanted it to be as generic as possible in case in future we wanted to allow props to configure a specific format

creido-welly: It can't be Date since that's a native JS object

creido-welly: Sorted

chriddell (approved):

PR #110, opened 2020-01-30
Feat/stylelint set order of content #6035

chriddell:
Resolves https://github.com/wellcometrust/corporate/issues/6035

- order of content:
  - `@include` (blockless)
  - `@extend`
  - custom-properties
  - declarations
  - `@include` (with block)
  - pseudo-states
  - stateful rules (e.g. &.is-active, &.has-state)
  - `@include mq` (media queries)
  - pseudo-elements (before/after)
  - rules (nested rules)
- forces new line before at-rules
  - except first nested at-rule
  - except blockless after blockless
  - ignores at-rule after a comment

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -61,6 +61,16 @@
                     "selector": "&:hover|&:focus|&:active|&:visited|&:focus-within"
                 },
                 {
+                    "type": "at-rule",
+                    "name": "include",
+                    "parameter": "hocus",
+                    "hasBlock": true
+                },
+                {
+                    "type": "at-rule",
+                    "hasBlock": true
+                },
+                {
                     "type": "rule",
                     "selector": "&.is-|&.has-|[[_a-zA-Z0-9].*]$"
                 },
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -51,18 +51,18 @@
                     "hasBlock": true
                 },
                 {
+                    "type": "at-rule",
+                    "name": "include",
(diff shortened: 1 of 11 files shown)


chriddell:
Full stylelint options if you wanted to test different scenarios: https://raw.githubusercontent.com/wellcometrust/corporate-components/edbca7ed7c4f4600a66a95c0720addedc467ecb8/.stylelintrc.json

src/components/Hamburger/_hamburger.scss
```diff
@@ -8,11 +8,11 @@
     color: var(--colour-white);
   }
 
-  &[disabled] {
+  @include mq(md) {
     display: none;
   }
 
-  @include mq(md) {
+  &[disabled] {
```
creido-welly, line 15: I reckon this qualifies as a state so should be added to the rule for .is-\.has

chriddell: 👍

.stylelintrc.json
```diff
@@ -18,16 +18,57 @@
...
                     "hasBlock": false
-                }
+                },
+                {
+                    "type": "at-rule",
+                    "name": "extend"
+                },
+                "custom-properties",
+                "declarations",
+                {
```
creido-welly, line 48: presumably this would include @include hocus so should that go underneath mq?

chriddell: good point, I would say that should live with pseudo-states

creido-welly (approved):

PR #115, opened 2020-01-31
#6045 bug: fix stylelint grid mixin order

chriddell:
Fixes https://github.com/wellcometrust/corporate/issues/6045

- removes default at-rule from stylelint rules
-  resets grid mixins to "original" state

### DIFF
--- a/src/components/Grid/styles/_20-mixins.scss
+++ b/src/components/Grid/styles/_20-mixins.scss
@@ -7,8 +7,6 @@
 // 2019-11-11
 // ----------------------------------
 
-/* stylelint-disable */
-
 /**
  * Sets up a grid container with a max-width and some margins
  */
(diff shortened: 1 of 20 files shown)


src/components/Grid/styles/_20-mixins.scss
```diff
@@ -280,6 +271,15 @@
       }
     }
   }
+  $autoflow: map-get($data, autoflow);
```
creido-welly: these mixins will still break as the variable declarations are being moved to bottom of the mixin block

chriddell: ah - the linter isn't actually catching them anymore, but this file is a hangover from prev commits. fixed

creido-welly: tnx

creido-welly (approved):

PR #123, opened 2020-02-12
#6111 feat(refactor): regulates use of react context

chriddell:
Resolves https://github.com/wellcometrust/corporate/issues/6111

- Moves _SearchPaneContext_ to component level

Ultimately we may want to manage Context differently (separate folder? within components?) but this brings all current uses of Context in to line with each other.


src/components/SearchPane/SearchPaneForm.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import Button from 'Button/Button';
+import Button from 'Button';
```
creido-welly, line 3: Surely this still needs to be `Button/Button` ?

chriddell: Nope, can be either - Button has an index.ts file so this works... however it does beg the question "which way is better?" (as both methods are used in this repo 😫).

I would argue the above is better, as it means if the main export in Button changes (to e.g. ButtonView) we can just update the Button/index.ts file to export that, and not have to update every component which uses the Button component

creido-welly: As discussed, if it's working in situ then all good. We can address the wider question of default vs named export in `src/index.ts` another time if need be

creido-welly (approved):

src/components/SearchPane/SearchPaneForm.tsx
chriddell: Will fix conflicts and merge today

PR #130, opened 2020-02-20
#6158 - feat: making sure created and updated date appears on certain…

DipakShrestha18:
… content types.

### DIFF
--- a/src/components/ResultsItem/ResultsItem.tsx
+++ b/src/components/ResultsItem/ResultsItem.tsx
@@ -39,18 +39,17 @@ export const ResultsItem = ({
           {meta.type}
           &nbsp;
         </span>
-        {meta.date && (nodeType === 'article' || nodeType === 'press_release') && (
+        {meta.date && ['article', 'press_release'].includes(nodeType) && (
           <span className="result-item__meta-date">
             <FormattedDate dateString={meta.date} />
             &nbsp;
           </span>
         )}
-        {meta.lastUpdated &&
-          (nodeType === 'article' || nodeType === 'press_release') && (
-            <span className="result-item__meta-last-updated">
-              Updated <FormattedDate dateString={meta.lastUpdated} />
-            </span>
-          )}
+        {meta.lastUpdated && ['article', 'press_release'].includes(nodeType) && (
+          <span className="result-item__meta-last-updated">
+            Updated <FormattedDate dateString={meta.lastUpdated} />
+          </span>
+        )}
       </div>
     )}
     <h3 className="result-item__title">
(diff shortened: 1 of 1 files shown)


src/components/ResultsItem/ResultsItem.tsx
```diff
@@ -19,6 +19,7 @@ export type ResultItemProps = {
   meta?: ResultItemMetaProps;
   text?: string;
   title: string;
+  nodeType?: string;
   type?: 'content' | 'file' | 'taxonomy_term';
```
chriddell, line 23: wondering if the naming of these two parameters is confusing now... what would `type` (content/file/taxonomy_term) be called in Drupal? File type? Content type?

src/components/ResultsItem/ResultsItem.tsx
```diff
@@ -37,15 +39,18 @@ export const ResultsItem = ({
...
           &nbsp;
         </span>
-        <span className="result-item__meta-date">
-          <FormattedDate dateString={meta.date} />
-          &nbsp;
-        </span>
-        {meta.lastUpdated && (
-          <span className="result-item__meta-last-updated">
-            Updated <FormattedDate dateString={meta.lastUpdated} />
+        {meta.date && (nodeType === 'article' || nodeType === 'press_release') && (
```
chriddell: How about:

```js
meta.data && ['article', 'press_release'].includes(nodeType)
```

instead? (more succinct, also more extensible)

src/components/ResultsItem/ResultsItem.tsx
```diff
@@ -37,15 +39,18 @@ export const ResultsItem = ({
...
+          <span className="result-item__meta-date">
+            <FormattedDate dateString={meta.date} />
+            &nbsp;
           </span>
         )}
+        {meta.lastUpdated &&
+          (nodeType === 'article' || nodeType === 'press_release') && (
```
chriddell: Same as above, how about:

```js
['article', 'press_release'].includes(nodeType)
```

src/components/ResultsItem/ResultsItem.tsx
DipakShrestha18: I'm trying to be really specific and trying to align with what you get from solr as well, that's why I called it nodeType.

chriddell: yeah with you there, wondering more about the name of the `type` parameter now that we have to "type" params...

chriddell: would it be better as `contentType`? or `fileType`?

chriddell: actually let's park this and address is another PR (if needs be)

chriddell (changes requested):
ignore type naming thing - see array.includes method only

src/components/ResultsItem/ResultsItem.tsx
DipakShrestha18: neat!! done.

chriddell (approved):

PR #132, opened 2020-02-21
#6170 feat(style): change footer/header bg color; adds X- prefix for …

chriddell:
…deprecated colours

- changes header/footer bg color to deprecated blue
- adds X- legacy prefix to vars

### DIFF
--- a/src/assets/styles/10-settings/_colours.scss
+++ b/src/assets/styles/10-settings/_colours.scss
@@ -39,6 +39,11 @@
 
   --text-colour: var(--colour-grey-60);
 
-  --colour-legacy-footer-error-red: #d2373f;
-  --colour-legacy-footer-error-pink: #ffebeb;
+  /**
+   * Use X- prefix to indicate deprecation. Try to avoid using any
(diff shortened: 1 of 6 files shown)


creido-welly (approved):
Have you checked the site logo background?

chriddell:
@creido good flag.

Ideally want to remove the background from the `.logo` selector, but noticed there are 3 logos there - the "medium" one is never visible, why is it there?

Screenshot 2020-02-25 at 09 48 35

creido-welly:
> Ideally want to remove the background from the `.logo` selector, but noticed there are 3 logos there - the "medium" one is never visible, why is it there?

Yes ideally the logo would have a transparent background so it wouldn't matter if the header colour behind it changed

Re the 3 logos, the first one is used on mobile devices when the user is at the top of the page. Scrolling down changes between this and the cut down version.

chriddell:
Of course! Thanks, couldn't tell what the third logo was doing there. Have updated the Header to hide the "medium" logo when non-sticky header, and removed bg.

PR #134, opened 2020-02-24
#6041 feat: add external link icon footer

chriddell:
Resolves https://github.com/wellcometrust/corporate/issues/6041

- adds Link component (detects external link)
- adds external link helper styles


src/components/Link/Link.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { useState, useEffect } from 'react';
```
creido-welly, line 1: We should integrate this functionality with RouterLinkWrapper which was setup in prep for routing injection.

chriddell: oh snap, I'll look into that

creido-welly: Granted RouterLinkWrapper doesn't leap out as 'Link' substitute :)

creido-welly (approved):

PR #135, opened 2020-02-25
Feat/corporate react cookie message #6168 (WIP)

creido-welly:
## In this PR
Cookie message markup and styles added from corporate repo to render basic component in prep for cookie logic work in the corporate-react app

## TODO
- refactor class names
- add style variations for buttons
- refactor styles


src/components/CookieMessage/_cookie-message.scss
```diff
@@ -0,0 +1,115 @@
+// TODO: conform styles to Wellcome Design System
```
chriddell, line 1: Presume these are pretty much copy-pasted from D7, with (spacing/colour) variables inserted?

chriddell (approved):
Looks good - those styles could do with looking at, but happy to tackle that in a separate PR as necessary

src/components/CookieMessage/_cookie-message.scss
creido-welly: yes indeed. I've literally just added some blurb about that in the description :)

creido-welly:
> Looks good - those styles could do with looking at, but happy to tackle that in a separate PR as necessary

Absolutely. I should've flagged up the fact that this is all WIP more clearly

PR #136, opened 2020-02-25
#6188 feat: configure stylelint to match corporate-react

chriddell:
See https://github.com/wellcometrust/corporate/issues/6188

### DIFF
--- a/src/assets/styles/20-tools/_mixins/_grid.scss
+++ b/src/assets/styles/20-tools/_mixins/_grid.scss
@@ -189,6 +189,11 @@
 
   $column-gap: false;
   $row-gap: false;
+  $column: 1;
+  $row: 1;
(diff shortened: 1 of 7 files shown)


src/assets/styles/20-tools/_mixins/_grid.scss
```diff
@@ -189,6 +189,11 @@
 
   $column-gap: false;
   $row-gap: false;
+  $column: 1;
+  $row: 1;
+  $counter: 1;
+
+  grid-auto-flow: $autoflow;
```
chriddell, line 196: FYI this is fine and doesn't break anything (grid-autoflow isn't currently used by anything, but I don't think it _will_ break if/when it is used)

creido-welly (approved):

PR #137, opened 2020-02-25
#6150 feat: adds + configs rollup-plugin-terser

chriddell:
Resolves https://github.com/wellcometrust/corporate/issues/6150

- adds rollup-plugin-terser to minify JS
- minifies CSS when production
- adds NODE_ENV variable to build task

### DIFF
--- a/config/rollup.config.js
+++ b/config/rollup.config.js
@@ -8,6 +8,7 @@ import autoprefixer from 'autoprefixer';
 import postcss from 'rollup-plugin-postcss';
 import cssvariables from 'postcss-css-variables';
 import calc from 'postcss-calc';
+import { terser } from 'rollup-plugin-terser';
 
 const extensions = [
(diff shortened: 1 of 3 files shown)


config/rollup.config.js
```diff
@@ -57,6 +58,7 @@ export default {
         ]
       }
     }),
+    terser(),
```
creido-welly: Presumably this will minify for dev and prod? Next step would be to apply it to prod only I guess using env variables

chriddell: done.

started looking at providing a sourcemap, but wouldn't this mean we'd need to start exporting the sources with the package?

think this may be a bigger discussion tbh... see conversation at https://github.com/googleapis/google-cloud-node/issues/2867

creido-welly (approved):
🚀

PR #139, opened 2020-02-26
#5940 feat: Button styles variations

creido-welly:


src/components/Button/Button.tsx
```diff
@@ -4,6 +4,7 @@ import cx from 'classnames';
 import Icon from 'Icon/Icon';
 
 type ButtonProps = {
+  buttonType?: 'primary' | 'secondary' | 'ghost' | 'link' | 'unstyled';
```
chriddell: Thoughts on re-naming this prop? 

```jsx
<Button buttonType="primary" />
```

is a bit verbose, thoughts on:

- `style`
- `appearance`
- `variation`

?

src/components/Button/_button.scss
```diff
@@ -1,16 +1,36 @@
...
   min-height: calc(5.5 * var(--space-unit));
   min-width: calc(5.5 * var(--space-unit));
-  padding: 0 calc(4 * var(--space-unit));
-  transition: background ease 0.4s;
+  padding: 0 calc(2.5 * var(--space-unit));
+  text-align: center;
+}
+
+.btn--primary {
+  background: var(--colour-blue-60);
```
chriddell: out of interest - any rationale behind using `background` and not `background-color`?

chriddell (approved):

src/components/Button/_button.scss
creido-welly: less to type is all

src/components/Button/Button.tsx
creido-welly: I thought best to avoid using _style_ as that's an existing HTML element attribute and _type_ would be disqualified for same reasons (and of course is already being used)

Appearance is same character length and variation is only 1 less :)

src/components/Button/_button.scss
chriddell: Read this recently and made me think we shouldn't use shorthand https://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/

(exceptions being margin & padding)

src/components/Button/Button.tsx
chriddell: no I mean verbose in that it repeats "button", if I'm passing any prop to a button I presume it's a prop for that button - ie. if it's `type` I presume "ah it's the button type", so if it's `buttonType` I'm thinking "obviously it's the *button*Type, it's for the button!"

tbh I don't hate `buttonType`, more just an inclination towards not repeating "button"

src/components/Button/_button.scss
creido-welly: fair point, updated

PR #141, opened 2020-02-26
#6190 bug: moves runtime-corejs to dependencies

chriddell:
Reopens and actually solves https://github.com/wellcometrust/corporate/issues/6190 

- moves `@babel/runtime-corejs3` from devDependencies to dependencies to fix issues with that package being missing when cc is imported

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.8.4",
+    "@babel/runtime-corejs3": "^7.8.4",
     "classnames": "^2.2.6",
     "date-fns": "^2.9.0",
     "lodash.throttle": "^4.1.1"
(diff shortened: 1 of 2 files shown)


annikaC (approved):

package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.8.4",
+    "@babel/runtime-corejs3": "^7.8.4",
```
creido-welly, line 48: prob best to full uninstall the package locally i.e. use `npm uninstall @babel/runtime-corejs3 --save-dev` and reinstall (as dependency) so that the change is registered in package-lock.json

chriddell: ah yes, will do

chriddell: done

PR #150, opened 2020-03-19
#6336 feat(style): removes text-transform from result items

chriddell:
Resolves https://github.com/wellcometrust/corporate/issues/6336


annikaC (approved):

DipakShrestha18:
Can you not transform only first letter at least until we find why solr is indexing lowercase? I think this will leave everything lowercased which something we do not want as well.

chriddell:
No sentence-case with CSS unfortunately. I'm thinking that it's "broken" whichever we do it, so better to display the data we're actually getting in the front-end rather than mutate it into something else whilst still not solving the root problem

PR #152, opened 2020-04-02
Feat/homepage header #6257 - Cross browser layout fixes

creido-welly:


src/assets/styles/10-settings/_typography.scss
```diff
@@ -17,18 +17,33 @@
...
+  // --text-regular:  16px
+  // --text-small:    14px
+  // --text-smaller:  13px
+  // --text-smallest: 11px
+  --text-larger: calc(20/16 * var(--text-base-size));
+  --text-large: calc(20/16 * var(--text-base-size));
+  --text-regular: var(--text-base-size);
+  --text-small: calc(14/16 * var(--text-base-size));
+  --text-smaller: calc(13/16 * var(--text-base-size));
+  --text-smallest: calc(11/16 * var(--text-base-size));
```
chriddell, line 46: considered naming these more explicitly? 

e.g. `--text-xxl-static` || `--text-xxl-fixed` etc

may be less confusing for someone who doesn't know the codebase?

chriddell (approved):

stuartwilkes (approved):

src/assets/styles/10-settings/_typography.scss
creido-welly: yeah it's a fair point although these are actually named after styles in the Wellcome Design System

we should discuss the bigger picture moving forward in terms of naming conventions for text

PR #156, opened 2020-04-02
#6257 homepage hero 

creido-welly:
- class name refactoring
- updated slideshow to cater for 1 or 0 items
- updated down arrow icon
- text styles

### DIFF
--- a/src/components/SlideshowHero/_slideshow-hero.scss
+++ b/src/components/SlideshowHero/_slideshow-hero.scss
@@ -28,12 +28,12 @@
   }
 }
 
-// TODO - rename class
-.grid-one {
+.slideshow-hero__content {
   @include grid-column(1, 7);
   
   @include mq($until: sm) {
     position: relative;
+    z-index: 1;
     
     // used as background to expand edges to reach to screen edges of parent grid
     &:before {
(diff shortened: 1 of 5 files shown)


src/components/SlideshowHero/_slideshow-hero.scss
```diff
@@ -70,19 +70,20 @@
   }
 }
 
-.slideshow-hero__content {
+.slideshow-hero__copy {
+  font-size: calc(18/16 * var(--text-base-size));
```
chriddell, line 74: why the 18/16?

chriddell (approved):

src/components/SlideshowHero/_slideshow-hero.scss
creido-welly: 18 pixels / the base pixels (in other words the calculation needed to get the decimal to multiply the text-base-size ... that wriggled and jiggled and tickled inside her)

chriddell: nice - seems like this could be a var from `_typography`, but I can also see that the design system declares two variants which are the same size (larger & large)... 😞

might be worth addressing this in the typography settings scss, and then we could have:

```css
--text-larger: calc(20/16 * var(--text-base-size));
--text-large: calc(18/16 * var(--text-base-size));
```

or something to that effect? seems redundant having larger and large the same size

PR #157, opened 2020-04-03
feat(refactor): global typography (WIP)

creido-welly:
- renamed global font size variables for better separation of non-responsive body text and heading styles based on Wellcome Design System


chriddell (approved):

chriddell (approved):
great, looks way tidier

PR #158, opened 2020-04-06
#6257 feat: home hero - spacing and IE tweaks

creido-welly:


src/components/SlideshowHero/_slideshow-hero.scss
```diff
@@ -89,6 +92,10 @@
   margin-bottom: var(--space-xl);
   transition: opacity ease var(--transition-duration);
   width: 100%;
+
+  @include mq($until: sm) {
+    margin-bottom: calc(2* var(--space-unit));
```
chriddell:
```css
2* var(--space-unit)
```

is the linter catching the space before the `*` and removing it? 

imo it would read better with spaces either side of the `*`

chriddell (approved):

src/components/SlideshowHero/_slideshow-hero.scss
creido-welly: I just omitted the space
def more readable with it
done
🎸

PR #163, opened 2020-04-20
Feat/home hero responsive images #6411

creido-welly:
- added switching of low-res to high-res image sources after Hero component has been initially rendered on client
- Data structure updated for flexibility i.e. an array of data will allow us to populate a variable number of `<source>` elements within a `<picture>` element
- responsive image markup extracted as separate Picture component

### DIFF
--- a/src/components/SlideshowHero/_slideshow-hero.scss
+++ b/src/components/SlideshowHero/_slideshow-hero.scss
@@ -194,10 +194,22 @@
 }
 
 .slideshow__image {
-  height: 100%;
-  object-fit: cover;
   position: absolute;
   width: 100%;
+
+  @supports (object-fit: cover) {
+    height: 100%;
+    object-fit: cover;
+  }
+
+  @media all and (-ms-high-contrast:none) {
+    top: 50%;
+    transform: translateY(-50%);
+  }
+
+  &.lazy {
+    filter: blur(20px) brightness(110%);
+  }
 }
 
 .slideshow__image-caption {
--- a/src/components/SlideshowHero/_slideshow-hero.scss
+++ b/src/components/SlideshowHero/_slideshow-hero.scss
@@ -206,10 +206,6 @@
     top: 50%;
     transform: translateY(-50%);
   }
-
-  &.u-loading-preload {
-    filter: blur(20px) brightness(110%);
-  }
 }
 
 .slideshow__image-caption {
--- a/src/components/SlideshowHero/_slideshow-hero.scss
+++ b/src/components/SlideshowHero/_slideshow-hero.scss
@@ -207,7 +207,7 @@
     transform: translateY(-50%);
   }
 
-  &.lazy {
+  &.u-loading-preload {
     filter: blur(20px) brightness(110%);
   }
 }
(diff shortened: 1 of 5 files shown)


src/components/SlideshowHero/_slideshow-hero.scss
```diff
@@ -194,10 +194,22 @@
...
+    height: 100%;
+    object-fit: cover;
+  }
+
+  @media all and (-ms-high-contrast:none) {
+    top: 50%;
+    transform: translateY(-50%);
+  }
+
+  &.lazy {
```
chriddell: Wondering about this classname... thinking should it be:

- `is-lazy` to denote a state... but it's not stateful really... so should it be
- `.slideshow-image--lazy` to denote modifier?

chriddell (changes requested):

src/components/SlideshowHero/_slideshow-hero.scss
creido-welly: `.u-lazy` prob better since (I hope) it'll become universal

thoughts?

lukewhitmore: Will a hero slideshow ever need lazy loading if it's always above the fold?

chriddell: Guess it's more "progressive" than lazy, in that we're loading a low-res placeholder to begin with to make the UI feel fast.

> .u-lazy 

Don't think it's a utility classname as it is, unless we want to change the name to `u-blur` to be more descriptive?

creido-welly: Agreed, it is above the fold so ideally shouldn't be called lazy loaded and it is indeed more progressive load.

Blur is more descriptive of the styling effect which gets applied whereas "preloading" is a better descriptor of what's happening technically. In which case how about one of the following?:

`.has-preload`
`.is-preload`
`.is-preloaded`

chriddell: the `has/is` would work for now, but it restricts it to only the `slideshow__image` selector (`.slideshow__image.is-preload`)

so if we wanted it to be universal selector it should be a utility class imo, and become something (e.g.) `u-img-preload`

creido-welly: sold!

src/components/SlideshowHero/_slideshow-hero.scss
```diff
@@ -194,10 +194,22 @@
...
+  }
+
+  @media all and (-ms-high-contrast:none) {
+    top: 50%;
+    transform: translateY(-50%);
+  }
+
+  &.u-loading-preload {
+    filter: blur(20px) brightness(110%);
+  }
```
chriddell: not sure about the nesting... how about adding the `u-loading-preload` selector to https://github.com/wellcometrust/corporate-components/tree/develop/src/assets/styles/50-utilities?

I think if the styles are specific to the `slideshow__image`, then use `&.is-loading-preload`, but if they're more agnostic then add to util partials

creido-welly: Done

chriddell: ace

chriddell (approved):

PR #166, opened 2020-04-22
Feat/add parent text data attribute #5708

stuartwilkes:
Adds a data-itempath attribute to nav items to allow GTM to read this with out having to traverse the DOM to generate it

### DIFF
--- a/src/components/Nav/NavItem.tsx
+++ b/src/components/Nav/NavItem.tsx
@@ -9,6 +9,7 @@ type Props = {
   className?: string;
   level: number;
   parentUrl?: string;
+  parentText?: string;
   text: string;
   url: string;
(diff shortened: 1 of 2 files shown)


creido-welly (approved):
@stuartwilkes this is probably how I would have done it so works for me.

Moving forward I'm thinking we need to look at how the nav links and sub nav links are populated so that the links for the current page and, perhaps more importantly, its parent are correctly set to active. My assumption is there's a menu API or similar in Drupal 8 from which we can get site map info and we would then query each node id against the data for the current page to mark the active. This may also change how we pass this info to GTM but we can cross that one when we come to it I think.

stuartwilkes:
@creido-welly Excellent - thank you.

PR #168, opened 2020-04-27
#6483 feature(style): tweaks cookie message styles

chriddell:
See https://github.com/wellcometrust/corporate/issues/6483

- tweaks cookie message styles

### DIFF
--- a/src/components/CookieMessage/_cookie-message.scss
+++ b/src/components/CookieMessage/_cookie-message.scss
@@ -4,12 +4,13 @@
   color: var(--colour-black);
   left: 0;
   letter-spacing: var(--body-letter-spacing);
-  padding: calc(4 * var(--space-unit)) calc(6 * var(--space-unit));
+  padding: var(--space-lg);
   position: fixed;
+  width: 100%;
   z-index: 105;
 
   @include mq(sm) {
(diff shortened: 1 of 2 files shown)


src/components/CookieMessage/_cookie-message.scss
```diff
@@ -4,12 +4,13 @@
   color: var(--colour-black);
   left: 0;
   letter-spacing: var(--body-letter-spacing);
-  padding: calc(4 * var(--space-unit)) calc(6 * var(--space-unit));
+  padding: var(--space-lg);
```
creido-welly: This padding is overridden on line 13, which is fine, but it's worth remembering that `--space-lg` will apply padding across the 3 main device widths

the CSS would be more efficient if line 7 was simply `padding: calc(2 * var(--space-unit));`

creido-welly (approved):
1 minor comment to address

src/components/CookieMessage/_cookie-message.scss
chriddell: very good point - I forgot the CSS vars are transpiled down, so thought the method I used wouldn't cause any issues!

creido-welly (approved):

PR #170, opened 2020-04-27
#6474 feat: Image component (WIP)

creido-welly:


src/components/Image/_image.scss
```diff
@@ -0,0 +1,30 @@
+// ----------------------------------
+// UI Components
+// Image
+// ----------------------------------
+// Design System 1.0 Alpha
+// 2019-11-11
+// ----------------------------------
+
+.cc-media {
```
chriddell, line 9: is the plan to update all the selectors to use this prefix?

chriddell (approved):
very tidy

src/components/Image/_image.scss
creido-welly: Still mulling that one over. It may not be necessary for majority of components in the ac.uk ecosystem but it does mean we have much more scope for including them elsewhere in future. Perhaps we can put this one to group vote

chriddell: It's a yes from me 😁

creido-welly: nuts to it, let's just do it moving forward

small PRs please 😆

PR #173, opened 2020-04-28
Feat/video component #6479

creido-welly:


src/components/Video/Video.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const [loaded, setLoaded] = useState(false);
+
+  const onLoad = () => {
+    setLoaded(true);
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const onError = (e: any) => {
+    console.error('Video error', e);
+  };
```
chriddell: This isn't used anywhere - presumably it's there for use in the future?

chriddell (approved):

src/components/Video/Video.tsx
creido-welly: Good point - I've removed this and onLoad.

It may be useful in future but since video will be embedded in an iframe the error handling can be done at the YouTube end.

chriddell (approved):

PR #185, opened 2020-06-02
#6729 feat: Generic <Listing /> wrapper component

creido-welly:


src/components/Listing/Listing.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import Listing from './Listing';
+
+describe('<Listing />', () => {
+  const output = shallow(
+    <Listing>
+      <li>List item 1</li>
+      <li>List item 2</li>
+      <li>List item 3</li>
+    </Listing>
```
chriddell: Small thing - but is this rendering 3 `<li>`s as children of a `<div>`?

chriddell (approved):
approved it, but minor comment on semantics in the Listing.test.tsx

src/components/Listing/Listing.test.tsx
creido-welly: Ha, good spot. The component was set to `ul` as  default originally hence the list items. Sorted

PR #186, opened 2020-06-03
#6667 feat: Gallery thumbnail view (WIP)

creido-welly:
- Lead image logic and styling
- Abstracted gallery images out as GalleryMedia component
- Abstracted img elements from Gallery and Image as ImageElement to allow for common setting of responsive attributes

### DIFF
--- a/src/components/Gallery/_gallery.scss
+++ b/src/components/Gallery/_gallery.scss
@@ -12,9 +12,12 @@
   margin: calc(-2 * var(--space-unit));
 }
 
-@mixin flex-grid-item($cols: 1) {
+%flex-grid-item {
   flex: 0 0 auto;
   padding: calc(2 * var(--space-unit));
+}
+
+@mixin flex-grid-item($cols: 1) {
   width: 100%/$cols;
 }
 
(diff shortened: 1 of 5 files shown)


src/components/Gallery/_gallery.scss
```diff
@@ -36,21 +39,38 @@
...
 }
 
 .cc-gallery__media-content {
   @extend %image-responsive;
 }
+
+.break {
+  flex-basis: 100%;
+  height: 0;
+}
```
chriddell, line 60: What does this do?

creido-welly: aha! magic (ssh)

alright then, it's a means of forcing flexbox items on to the next row. In this instance I'm using it to ensure the lead image (if there is one) is the only item on its row. Not the most helpfully named class right now I'll admit - how about `u-flexgrid-break`?

chriddell: ah okay - would `flex-wrap: wrap` on the parent work, instead?

creido-welly: sadly not - as discussed this does appear to be the neatest solution to the problem

chriddell: agreed

chriddell (approved):

PR #188, opened 2020-06-05
#6743 feat: Gallery thumbnail view - image data

creido-welly:
- passed onClick handler from parent Gallery component to child images which are now being passed in as children as opposed to the image data coming in via props
- added image alt text
- image interaction styles

### DIFF
--- a/src/components/Gallery/Gallery.tsx
+++ b/src/components/Gallery/Gallery.tsx
@@ -1,10 +1,11 @@
-import React, { MouseEventHandler } from 'react';
+import React, { Children, cloneElement, MouseEventHandler } from 'react';
 import cx from 'classnames';
 
 import Button from 'Button';
 import ImageElement from 'Image/ImageElement';
 
 type GalleryMediaProps = {
+  alt?: string;
   height: number;
   isLead?: boolean;
   onClick: MouseEventHandler;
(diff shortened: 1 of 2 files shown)


src/components/Gallery/Gallery.tsx
```diff
@@ -29,29 +31,43 @@ const GalleryMedia = ({
...
   const handleOnClick = (e: any) => {
-    console.log(e.currentTarget);
+    // TODO open gallery lightbox
+    console.log('Gallery handleOnClick', e.currentTarget);
   };
 
+  const childrenWithProps = Children.map(children, child =>
+    cloneElement(child, { onClick: handleOnClick })
+  );
+
```
chriddell, line 65: 👌

src/components/Gallery/_gallery.scss
```diff
@@ -41,17 +41,56 @@
...
+  appearance: none;
+  background: var(--colour-grey-50);
+  border: 0;
+  color: var(--colour-white);
+  cursor: pointer;
+  display: block;
+  height: 100%;
+  outline: 0;
+  position: relative;
+  width: 100%;
```
chriddell, line 60: Presuming some of this is resetting `<button>` styles? I'd probably abstract those styles into some kind of btn-reset mixin/extend/util to make it re-usable, but not hugely important

chriddell (approved):
left a comment re: `button` style abstraction, but shouldn't prevent a merge if you don't want it to!

src/components/Gallery/_gallery.scss
creido-welly: indeed :)

PR #190, opened 2020-06-09
#6681 - in page navigation component.

DipakShrestha18:
Styling still to be done.


src/components/InpageNav/InpageNav.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+type InpageNavItemProps = {
+  title?: string;
+  id?: string;
+};
+
+export const InpageNav = ({ links }: InpageNavItemLinksProps) => {
+  return (
+    <>
+      <h3 className="section-links__heading">On this page</h3>
```
creido-welly: I think a `<strong>` tag would prob be more appropriate for this than `<h3>`. @chriddell your thoughts?

src/components/InpageNav/InpageNav.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <ul className="list-unstyled list-highlight">
+        {links.map(({ title, id, ...props }) => (
+          <li className="list-highlight__item">
+            <p className="list-highlight__title">
```
creido-welly: We can def get rid of the paragraph tag wrapping the <a>

DipakShrestha18: what should it be instead?

src/components/InpageNav/InpageNav.tsx
```diff
@@ -0,0 +1,29 @@
...
+  links?: InpageNavItemProps[];
+};
+
+type InpageNavItemProps = {
+  title?: string;
+  id?: string;
+};
+
+export const InpageNav = ({ links }: InpageNavItemLinksProps) => {
+  return (
```
creido-welly: Minor: this works but from a tidiness PoV, as there are no `const` declarations or other javascript going on in this function you could simplify with implicit return i.e.
```jsx
export const InpageNav = ({ links }: InpageNavItemLinksProps) => (
  {/* JSX goes here */}
)
```

src/components/InpageNav/InpageNav.tsx
creido-welly: the `<li>` wrapping the `<a>` is all we need :)

src/components/InpageNav/InpageNav.tsx
DipakShrestha18: Do you think this will require class as well.` <strong class"">`

creido-welly: It will need a class yes (`className` in React)  and that should stay the same. The strong tag simply implies more importance than standard text and although that would usually mean it was styled bold that should have nothing to do with how it actually looks

DipakShrestha18:
@creido-welly made changes as suggested. Styling still needs to be done.

creido-welly (approved):

PR #191, opened 2020-06-12
#6670 feature: add gallery lightbox view

chriddell:
- adds a view for the GalleryLightbox

### What this PR doesn't do:
- Integrate the GalleryLightBox into Gallery (see https://github.com/wellcometrust/corporate/issues/6673)
- Doesn't include social share (see https://github.com/wellcometrust/corporate/issues/6671)
- Not possible to hide/show the info-pane (see https://github.com/wellcometrust/corporate/issues/6674)

Also - the layout is not perfectly matching of those at https://app.zeplin.io/project/5b4e06b48ae4580d4871178a/dashboard?q=gallery - this is due to some fairly impractical (if not impossible) layout choices in the designs. I'm hoping to merge this view as is so that we can create a feature branch from corporate-react which uses it, to get designer feedback.


chriddell:
Not quite ready to merge, needs chat with @j-jaworski re: some tricky layout.

src/components/GalleryLightBox/_gallery-lightbox.scss
```diff
@@ -0,0 +1,207 @@
...
+  font-size: var(--body-xs);
+  margin-bottom: var(--space-md);
+  padding-left: var(--space-md);
+}
+
+.cc-gallery-lightbox__meta-item {
+  display: inline;
+
+  // Override styles from wysiwyg
+  > p {
```
creido-welly, line 161: wysiwyg overrides are an issue for a number of components.

We don't need to fix it across the board in a perfect generic/future proof/#componentLibrariesWillNeverBeBetterThanThis fashion right now - it's something for the future

creido-welly (approved):
One minor comment, otherwise 🚀

src/components/GalleryLightBox/_gallery-lightbox.scss
chriddell: ah - so you're saying I should override this from within corporate-react as opposed to here?

creido-welly: No I think corporate-components IS the place to override it.

I've done a similar thing for Text component. There are also text snippets to think about (very like Text paragraph type but much more pared back editing capabilities - basically a Text paragraph for an infobox).

Could this be an application for an object? e.g. `.o-rich-text` as it seems more than just utility.

Like I said, not something to address now because we have more pressing things to deliver first.

chriddell: ah I get you, I think it would be place for adding a "scoped" selector as I think it falls more under that category than an object https://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/#scope-namespaces-s-

PR #193, opened 2020-06-16
#6827 feat: Added SocialShare component

creido-welly:
* added component (simplification needed)
* updated social icons to match latest design


chriddell (approved):
Looks good to me

PR #196, opened 2020-06-18
#6821 feat: ImageBanner component

creido-welly:
Work in progress ...

### DIFF
--- a/src/components/ImageBanner/ImageBanner.tsx
+++ b/src/components/ImageBanner/ImageBanner.tsx
@@ -30,16 +30,14 @@ export const ImageBanner = ({
   });
 
   return (
-    <figure className={classNames}>
-      <div className="cc-image-banner__container" role="banner">
-        <ImageElement
-          alt={alt}
-          className="cc-image-banner__img"
-          src={src}
-          role="presentation"
-          sizes={sizes}
-        />
-      </div>
+    <figure className={classNames} role="banner">
+      <ImageElement
+        alt={alt}
+        className="cc-image-banner__img"
+        src={src}
+        role="presentation"
+        sizes={sizes}
+      />
 
       <div className="cc-image-banner__caption-container">
         <MediaCaption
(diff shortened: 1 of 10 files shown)


src/components/ImageBanner/ImageBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+  src,
+  srcSet
+}: ImageBannerProps) => {
+  const classNames = cx('cc-image-banner', {
+    [`${className}`]: className
+  });
+
+  return (
+    <figure className={classNames}>
+      <div className="cc-image-banner__container" role="banner">
```
lukewhitmore: Could this div and the previous figure be combined to simplify?

src/components/ImageBanner/ImageBanner.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <ImageElement
+          alt={alt}
+          className="cc-image-banner__img"
+          src={src}
+          role="presentation"
+          sizes={sizes}
+        />
+      </div>
+
+      <div className="cc-image-banner__caption-container">
```
lukewhitmore, line 42: Should this div be a figcaption element?

src/components/ImageBanner/ImageBanner.tsx
lukewhitmore: If they can't be simplified .. perhaps ARIA role should be on figure anyway? Might make more sense?

src/components/ImageBanner/ImageBanner.tsx
creido-welly: Good spot, MediaCaption component which this wraps is actually a figcaption

src/components/ImageBanner/ImageBanner.tsx
creido-welly: It looks like they could be just looking at it briefly so I've made this change now.

This is largely a copy and paste from Drupal 7 right now as it's WIP ... more to follow

lukewhitmore (approved):

PR #197, opened 2020-06-18
#6674 feature: adds GalleryLightBox info pane

chriddell:
- adds info pane for GalleryLightBox
- ability to open/close (setState) from button onClick
- adds some dummy content to StoryBook for preview

### DIFF
--- a/src/components/GalleryLightBox/GalleryLightBox.tsx
+++ b/src/components/GalleryLightBox/GalleryLightBox.tsx
@@ -1,4 +1,5 @@
-import React, { useState } from 'react';
+import React, { useState, useContext } from 'react';
+import cx from 'classnames';
 
 import {
   CarouselProvider,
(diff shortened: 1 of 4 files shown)


src/components/GalleryLightBox/GalleryLightBox.tsx
```diff
@@ -10,8 +11,12 @@ import {
 
 import ImageElement from 'Image/ImageElement';
 import Icon from 'Icon';
+import Button from 'Button';
+
+import ViewportContext from 'ViewportContext/ViewportContext';
```
creido-welly, line 16: YESSSSSS! It's being used outside of the header 😄

creido-welly (approved):

src/components/GalleryLightBox/GalleryLightBox.tsx
chriddell: it's a __very__ handy piece of context

PR #202, opened 2020-06-19
#6819 - include style, update microcopy

DipakShrestha18:


chriddell (approved):
syntax all looks fine.

one point, though, is that it would be good if the PR had a description to let me know what's happening (and why!) - for example in this PR, all of the consent (handleConsentChange, setConsent etc) functions have disappeared, but I have no idea why! 🚀

DipakShrestha18:
> syntax all looks fine.
> 
> one point, though, is that it would be good if the PR had a description to let me know what's happening (and why!) - for example in this PR, all of the consent (handleConsentChange, setConsent etc) functions have disappeared, but I have no idea why! 🚀

Good point! I would have though if you would like to know more details, you can always go to the ticket #6819 which is appended in branch name and get the details.

chriddell:
That's true and is, after all, why we include the ticket # in the branch. 

I think with us working on different projects it helps to smooth the code review process if there's a description written which describes just the PR - not necessarily the entire issue that it relates to (as that may be much bigger than the PR!), though _some_ context is obviously necessary, potentially by linking the Github ticket from the description?

PR #207, opened 2020-06-23
#6673 feature: integrates GalleryLightbox

chriddell:
- adds click handler to GalleryMedia to open lightbox
- adds use-react-hotkeys package 
- useHotkeys hook (from use-react-hotkeys) to close GalleryLightbox on esc
- adds autoFocus prop to Button https://www.w3schools.com/tags/att_autofocus.asp
- deletes GalleryLightbox test & stories* (not currently used and they throw linting errors)

*Ticket for creating storybook component https://github.com/wellcometrust/corporate/issues/6844


src/components/GalleryLightBox/GalleryLightBox.tsx
```diff
@@ -80,15 +96,38 @@ export const GalleryLightBox = ({ slides, isOpen }: GalleryLightBoxProps) => {
...
+   * Use react-hot-keys hook to detect user clicking 'escape'
+   * to close the lightbox.
+   */
+  useHotkeys('esc', () => {
+    handleClose();
+  });
+
   return (
-    <dialog className="cc-gallery-lightbox" open={isOpen}>
+    <div
```
creido-welly: Out of interest why have you opted for `<div role="dialog">` instead of `<dialog>`?

src/components/GalleryLightBox/GalleryLightBox.tsx
```diff
@@ -80,15 +96,38 @@ export const GalleryLightBox = ({ slides, isOpen }: GalleryLightBoxProps) => {
...
       <CarouselProvider
         className="cc-gallery-lightbox__carousel"
         dragEnabled={false}
+        isIntrinsicHeight
         naturalSlideWidth={16}
         naturalSlideHeight={9}
         totalSlides={slides.length}
       >
+        <Button
+          autoFocus
```
creido-welly, line 117: 👍  nice

creido-welly (approved):
Good work, just one minor query

src/components/GalleryLightBox/GalleryLightBox.tsx
chriddell: it's a good question - two reasons (though I may actually change it back to a `<dialog>` now you've made me think about it...)

- it's not supported in IE (though I don't know how it falls back...)
- I was having a lot of trouble capturing the focus and handling a keypress, though I've solved that with the useHotkeys hook, and I thought it may be some behaviour in the `<dialog>`*

I didn't figure out the issues with this, so I'm going to make sure there's a valid reason to remove the dialog, I've a feeling it may be able to go back in so long as IE falls back well...

creido-welly (approved):

PR #214, opened 2020-07-06
Feat/gallery social share #6671

creido-welly:
- added social share icons including link sharing to lightbox
- tailored styles for lightbox colour scheme


lukewhitmore (approved):
Do social cards need to be thought about as part of this? (open graph / schema.org / microdata)

creido-welly:
**og** and **twitter** meta data is already in place but no `og:image` or `twitter:image` as yet. Now we have listing images in the data we can add those in - I'll create a ticket for that shortly

Regarding schema/microdata there is none in place on live WPP. What did you have in mind?

DipakShrestha18:
I think we should add listing image for og:image and twitter:image. I don't think we need to create another image for this.

creido-welly:
> I think we should add listing image for og:image and twitter:image. I don't think we need to create another image for this.

@DipakShrestha18 I agree (if I'm reading you correctly), there shouldn't be any backend work needed since we already have listing image in the data which we can use to populate OG and twitter images. I've created #6958 for the frontend work.

PR #222, opened 2020-07-17
Feat/gallery lightbox design feedback #6674

creido-welly:
Corrections to layout based on design feedback


src/components/GalleryLightBox/GalleryLightBox.tsx
```diff
@@ -81,6 +86,15 @@ export const GalleryLightBox = ({
```
creido-welly, line 86: Changes from this point onwards are largely moving things around as opposed to massive markup overhaul

stuartwilkes (approved):
🎉

PR #226, opened 2020-07-21
Fix/ie11 #7000

creido-welly:
#7000 fix: IE11 - debugging and gallery lightbox styling

* isolated errors caused by query-string and react-hotkeys-hook dependencies
* added replacement location.hash parsing
* added basic IE11-only styles

#7022 feat: Set download links to open images in a new tab


src/components/GalleryLightBox/GalleryLightBox.tsx
```diff
@@ -199,9 +199,11 @@ export const GalleryLightBox = ({
                     )}
                     <div className="cc-gallery-lightbox__download">
                       <a
-                        href={slide.mediaSources.gallery_full_hi}
-                        download
                         className="cc-gallery-lightbox__download-link u-color-inherit"
+                        download
```
lukewhitmore, line 203: I think download only works with same origin .. not sure what would happen in this case, re. headless?

creido-welly: yeah you're right it doesn't work in our case but adding it in hopes we can weave some magic in future :)

creido-welly: target blank added to open image in a new tab

PR #229, opened 2020-07-21
Corrected spelling and update link inline with current notice on D7 s…

lukewhitmore:
…ite. Also added in aria-label text to bring markup inline with D7 solution.

### DIFF
--- a/src/components/NewsletterForm/NewsletterFormFooter.tsx
+++ b/src/components/NewsletterForm/NewsletterFormFooter.tsx
@@ -5,12 +5,19 @@ import NewsletterFormItem from './NewsletterFormItem';
 export const NewsletterFormFooter = () => (
   <>
     <p className="newsletter-form__item newsletter-form__item--footer">
-      By clicking subscribe, you agree to recieve this newsletter.
+      By clicking subscribe, you agree to receive this newsletter.
     </p>
     <p className="newsletter-form__item newsletter-form__item--footer">
       We use a third party provider, Dotmailer, to deliver our newsletters. For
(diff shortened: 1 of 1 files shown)


DipakShrestha18 (approved):

src/components/NewsletterForm/NewsletterFormFooter.tsx
```diff
@@ -5,12 +5,19 @@ import NewsletterFormItem from './NewsletterFormItem';
...
   <>
     <p className="newsletter-form__item newsletter-form__item--footer">
-      By clicking subscribe, you agree to recieve this newsletter.
+      By clicking subscribe, you agree to receive this newsletter.
     </p>
     <p className="newsletter-form__item newsletter-form__item--footer">
       We use a third party provider, Dotmailer, to deliver our newsletters. For
-      information about how we handle your data, please read our privacy notice.
-      You can unsubscribe at any time using the links in the email you recieve.
+      information about how we handle your data, please read our{' '}
```
creido-welly, line 12: What's with the `{' '}`?

Is that instead of `&nbsp;`?

creido-welly (approved):

src/components/NewsletterForm/NewsletterFormFooter.tsx
lukewhitmore: The code sniffer suggested it  .. on retrospect, figured it's quite a good way of explicitly showing we want a space .. `&nbsp;` isn't technically the right solution I guess, as we actually only require a regular one?

PR #233, opened 2020-07-22
update id of newsletter signup.

DipakShrestha18:


src/components/InpageNav/InpageNav.stories.tsx
```diff
@@ -30,7 +30,7 @@ const InpageNavExample = () => {
   return <InpageNav isMinimal={isMinimal} links={links} />;
 };
 
-const stories = storiesOf('Components|Page header', module);
+const stories = storiesOf('Components|InPageNav', module);
```
creido-welly: This was grouped together with the Page header modules for easy author reference

src/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -26,10 +27,7 @@ export const NewsletterSignup = ({
   <Section className="newsletter-signup">
     <Grid>
       <GridCell column={1} columnCount={1}>
-        <h3
-          className="newsletter-signup__title"
-          id={title.replace(/\s+/g, '-').toLowerCase()}
-        >
+        <h3 className="newsletter-signup__title" id={inpageId}>
```
creido-welly: I presume the purpose of this to allow users to access newsletter signup from inpage nav?

If so you should take a look at section title to see how header overlap is dealt with (basically `.cc-section__title-anchor` positions a child span above the title to effectively position the title text itself underneath the header)
https://github.com/wellcometrust/corporate-components/blob/98259b8cd4211b44ce569fec13c5fd6065000542/src/components/SectionTitle/SectionTitle.tsx#L18

DipakShrestha18: It's not for inpage nav, just need to have id so it can be linked from links within the page text.

src/components/InpageNav/InpageNav.stories.tsx
DipakShrestha18: Reverted.

src/components/NewsletterSignup/NewsletterSignup.tsx
creido-welly: ok

the same would apply though I think, since this is all based around standard HTML anchor functionality i.e. the newsletter signup title could be covered up by the header nav

DipakShrestha18: I'm bit concerned about what affect will it have on styling when I change h3 to span.

creido-welly: NO, don't do that

The fix is to insert a child span _inside_ the H3

creido-welly: SectionTitle markup in full - https://github.com/wellcometrust/corporate-components/blob/98259b8cd4211b44ce569fec13c5fd6065000542/src/components/SectionTitle/SectionTitle.tsx#L16-L21

DipakShrestha18: I see, ok i have made an update.

creido-welly (approved):

PR #234, opened 2020-07-22
remove content type from search result as it's not reflect true type.

DipakShrestha18:


src/components/ResultsItem/ResultsItem.tsx
```diff
@@ -34,12 +34,14 @@ export const ResultsItem = ({
   <article className="result-item">
     {meta && (meta.type || meta.date) && (
       <div className="result-item__meta">
-        {meta.type && !['taxonomy_term'].includes(itemType) && (
+        {/* remove this for now as types are not correctly displayed in search result
```
creido-welly, line 37: Minor: Fair enough - t if you have time , would be useful to create a ticket and add the ticket number in this comment for reference

creido-welly (approved):

src/components/ResultsItem/ResultsItem.tsx
DipakShrestha18: will do it.

PR #246, opened 2020-08-10
#7133 feat: Gallery thumbnail view - prevent orphans

creido-welly:
This PR adds logic to insert a row break which will force the 2nd to last gallery item on to the next row to prevent any orphans

### DIFF
--- a/src/components/Gallery/Gallery.tsx
+++ b/src/components/Gallery/Gallery.tsx
@@ -21,6 +21,7 @@ type GalleryMediaProps = {
   fileSize?: number;
   height: number;
   isLead?: boolean;
+  isNewRow?: boolean;
   licence?: string;
   mediaSources: any;
   onClick?: MouseEventHandler;
@@ -56,6 +57,7 @@ export const GalleryMedia = ({
   fileSize,
   height,
   isLead,
+  isNewRow,
   licence,
   mediaSources,
   onClick = () => {},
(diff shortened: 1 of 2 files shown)


src/components/Gallery/Gallery.tsx
```diff
@@ -103,7 +106,7 @@ export const GalleryMedia = ({
           />
         </Button>
       </div>
-      {isLead && <div className="break" />}
```
stuartwilkes, line 106: Just checking - is this change for style consistency, or a technical reason? Just wondering!

stuartwilkes (approved):

lukewhitmore:
Just out of interest, could this have been carried out purely using flexbox? What are the benefits of using the extra element?

src/components/Gallery/Gallery.tsx
creido-welly: It is (unfortunately) purely for layout/styling reasons. Loathe as I am to use non-semantic styles and coupled CSS+HTML, @chriddell and I talked thru options at the time and realised that this was really the only way to make row breaks work

creido-welly:
> Just out of interest, could this have been carried out purely using flexbox? What are the benefits of using the extra element?

@lukewhitmore unfortunately not (I tried 😢 ) https://github.com/wellcometrust/corporate-components/pull/246#discussion_r468453832

src/components/Gallery/Gallery.tsx
stuartwilkes: Fair enough - there's always cases where the non-idealistic solution is the best option...

PR #253, opened 2020-08-27
#7173 feature: adds SiteAlert component + styles

chriddell:
### Context

We need to utilise what is now highly specific piece of UI (the WellcomeCollectionBanner) for another purpose, so we may as well create a much more agnostic component (a SiteAlert) which can be used for a multitude of purposes in the same context. This PR creates that component.

See https://github.com/wellcometrust/corporate/issues/7173 for more.

- adds SiteAlert component (in prep for moving WellcomeCollectionBanner)
- adds styles
- adds test file
- adds storybook file

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -32,6 +32,7 @@ export { SearchForm } from 'SearchForm/SearchForm';
 export { SearchPane } from 'SearchPane/SearchPane';
 export { Section } from 'Section/Section';
 export { SectionTitle } from 'SectionTitle/SectionTitle';
+export { default as SiteAlert } from 'SiteAlert';
 export { SlideshowHero } from 'SlideshowHero/SlideshowHero';
 export { SocialShare } from 'SocialShare/SocialShare';
 export { Text } from 'Text/Text';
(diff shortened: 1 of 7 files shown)


src/index.ts
```diff
@@ -32,6 +32,7 @@ export { SearchForm } from 'SearchForm/SearchForm';
 export { SearchPane } from 'SearchPane/SearchPane';
 export { Section } from 'Section/Section';
 export { SectionTitle } from 'SectionTitle/SectionTitle';
+export { default as SiteAlert } from 'SiteAlert';
```
chriddell, line 35: This is a slightly different method of exporting, but I think it's preferable and the other exports in this file should follow the same pattern:

- makes easier to change components' main export without touching this file (e.g. if SiteAlert/index.ts wants to `export default SiteAlertView` rather than `export default SiteAlert`, then we would not need to touch this file 🚀 
- prevents a linting error on the SiteAlert/index.ts because the export is never used (not imported elsewhere in the app)

DipakShrestha18 (approved):

PR #255, opened 2020-09-02
#7187 bug: Card component view logic

creido-welly:
* Added view logic to ensure vertical and horizontal variants present data correctly
* ResultsItem - corrected HTML parsing

### DIFF
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -17,6 +17,7 @@ export type CardImageProps = {
 
 export type CardProps = CardBaseProps & {
   image?: CardImageProps;
+  variant: 'horizontal_card' | 'link_list' | 'text_list' | 'vertical_card';
 };
 
 const cardImageSizesDefault =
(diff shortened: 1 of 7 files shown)


src/components/Card/Card.tsx
```diff
@@ -72,24 +79,28 @@ export const Card = ({
...
+              <dl className="cc-card__authors">
+                <dt className="cc-card__authors-label">Author</dt>
+                {authors?.map(author => (
+                  <dd key={author} className="cc-card__author">
+                    {author}
+                  </dd>
+                ))}
+              </dl>
+            ) : (
+              <span />
```
chriddell: How come the empty span tag?

chriddell (approved):

src/components/Card/Card.tsx
creido-welly: `justify-content: space-between` 😬 

prob should change the styling really

creido-welly: Much better now

PR #259, opened 2020-09-07
#7228 feature: adds DateInput component + export

chriddell:
Relates to https://github.com/wellcometrust/corporate/issues/7228

- adds DateInput component + export

### DIFF
--- a/src/components/DateInput/DateInput.tsx
+++ b/src/components/DateInput/DateInput.tsx
@@ -3,10 +3,10 @@ import cx from 'classnames';
 
 type DateInputProps = {
   className?: string;
-  name: string;
   id: string;
   isInvalid?: boolean;
   isRequired?: boolean;
+  name: string;
 };
 
 export const DateInput = forwardRef(
--- a/src/components/DateInput/DateInput.tsx
+++ b/src/components/DateInput/DateInput.tsx
@@ -1,0 +1,35 @@
+import React, { forwardRef } from 'react';
+import cx from 'classnames';
+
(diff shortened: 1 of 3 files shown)


creido-welly:
This seems to be almost a carbon copy of TextInput except for type="date". Is there a specific reason why they are 2 separate components?

RadioInput also seems to be extremely close - could they be amalgamated into one and the type injected as a prop (which could also define the className)?

If there's a good reason then fair enough

chriddell:
It's a good point, and one I did consider... not sure if this is a "good" reason, but I found working on the PoC that once these Input components start to expand they become quite hard to maintain as a single component. When we started to add features like field validation, error messaging and styles it became quite unwieldy having all of this in one file (particularly with the Date field). 

That said, I'm not hugely wedded to the idea of maintaining separate components but it is my preferred route - what do you think @creido-welly?

*edit: also - my idea with the separate components style-wise is to use an `@extend input-base` placeholder, so that we don't have to maintain repeated styles between components

creido-welly:
@chriddell I think it's fine for now as long as you have a good rationale behind why you're doing it. We can keep an eye on and if consolidating it helps in future then we can do that.

:+1: for the style `@extends`

creido-welly (approved):

chriddell:
I think once I start using these components to build a working form, we'll start to get a better picture of whether we can combine the three components or whether it makes sense to keep apart. 

Will keep an eye on how these develop and address if needs be.

PR #264, opened 2020-09-07
Feat/tableau #7195

creido-welly:
This PR includes the TableauChart component and a consolidation of file download markup with ResultsItem (used in Search and Text listings)

### DIFF
--- a/src/components/FileDownload/FileDownload.tsx
+++ b/src/components/FileDownload/FileDownload.tsx
@@ -19,7 +19,7 @@ export const FileDownload = ({
   type
 }: FileDownloadProps) => {
   const classNames = cx('cc-file-download', {
-    [`${className}`]: className
+    [className]: className
   });
 
   return (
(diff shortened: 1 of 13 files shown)


src/components/FileDownload/FileDownload.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const FileDownload = ({
+  className,
+  href,
+  label = 'Download',
+  name,
+  size,
+  type
+}: FileDownloadProps) => {
+  const classNames = cx('cc-file-download', {
+    [`${className}`]: className
```
chriddell: Template literal/placeholder not actually needed in this instance - you can initialise as just `[className]` and all still works - much of a muchness though 🙃

chriddell (approved):

src/components/FileDownload/FileDownload.tsx
creido-welly: Yeah copied and pasted for the umpteenth time 😬 

Removed it here but I'll do all of the other 29 instances in a separate PR 😆

PR #266, opened 2020-09-08
#7241 feature: adds label prop to RadioInput

chriddell:
Relates to https://github.com/wellcometrust/corporate/issues/7241

- adds `label` prop to RadioInput
- refactors RadioInput to allow conditional rendering based on `label` prop being present

### DIFF
--- a/src/components/RadioInput/RadioInput.tsx
+++ b/src/components/RadioInput/RadioInput.tsx
@@ -46,7 +46,7 @@ export const RadioInput = forwardRef(
       </div>
     ) : (
       <input
-        className={cx({ [className]: className })}
+        className={className ? cx({ [className]: className }) : undefined}
         id={id}
         name={name}
         ref={ref}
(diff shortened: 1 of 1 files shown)


src/components/RadioInput/RadioInput.tsx
```diff
@@ -1,28 +1,52 @@
...
+          ref={ref}
+          required={isRequired}
+          type="radio"
+          value={value}
+        />
+      </div>
+    ) : (
       <input
-        className={classNames}
+        className={cx({ [className]: className })}
```
chriddell: Working on the presumption here that if a RadioInput is used without a `label` prop passed, the use-case will be highly specific and therefore we don't render with a "default" classname

src/components/RadioInput/RadioInput.tsx
```diff
@@ -1,28 +1,52 @@
...
+          ref={ref}
+          required={isRequired}
+          type="radio"
+          value={value}
+        />
+      </div>
+    ) : (
       <input
-        className={classNames}
+        className={className ? cx({ [className]: className }) : undefined}
```
creido-welly: Minor: setting `null` instead of `undefined` would be more inline with React convention

creido-welly (approved):

PR #267, opened 2020-09-08
#7242 feature: adds TextArea component + export

chriddell:
Relates to https://github.com/wellcometrust/corporate/issues/7242

- adds TextArea component + export

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -42,6 +42,7 @@ export { SlideshowHero } from 'SlideshowHero/SlideshowHero';
 export { SocialShare } from 'SocialShare/SocialShare';
 export { TableauChart } from 'TableauChart/TableauChart';
 export { Text } from 'Text/Text';
+export { default as TextArea } from 'TextArea';
 export { default as TextInput } from 'TextInput';
 export { Video } from 'Video/Video';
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -42,6 +42,7 @@ export { SlideshowHero } from 'SlideshowHero/SlideshowHero';
 export { SocialShare } from 'SocialShare/SocialShare';
 export { TableauChart } from 'TableauChart/TableauChart';
 export { Text } from 'Text/Text';
+export { default as TextArea } from 'TextArea';
```
creido-welly, line 45: Just out of interest, why have you done it this way instead of `... from TextArea/TextArea`?

creido-welly (approved):

src/index.ts
chriddell: _(copy and pasted the below from an earlier PR where I'd done the same thing - https://github.com/wellcometrust/corporate-components/pull/253)_

This is a slightly different method of exporting, but I think it's preferable and the other exports in this file should follow the same pattern:

- makes easier to change components' main export without touching this file (e.g. if SiteAlert/index.ts wants to `export default SiteAlertView` rather than `export default SiteAlert`, then we would not need to touch this file (src/index.ts) 🚀 
- prevents a linting error on the SiteAlert/index.ts because the export is never used (not imported elsewhere in the app)

PR #274, opened 2020-09-09
#7256 feat: New window links in rich text (WIP)

creido-welly:
Work in progress

* Added RichText component to handle processing of rich text specifically to insert new window icons and assistive text where needed
* Polyfill added for String.prototype.replaceAll()
* Marked existing component files to which new window link icons are to be added


src/components/RichText/_text.scss
```diff
@@ -0,0 +1,80 @@
+// ----------------------------------
+// UI Components
+// Text | TextSnippet
+// ----------------------------------
+// Design System 1.0 Alpha
+// 2019-11-11
+// ----------------------------------
+
+.cc-text {
```
chriddell: I created [this file](https://github.com/wellcometrust/corporate-components/pull/273/files#diff-c8dfa52107486fa946af088a312fae3b) as part of a previous PR.

I created it to cover style issues where we don't (or in future _wont_) have control over the markup (e.g. from an RTE).

I feel like we could drop the file I created and use the RichText component you've created to render all instances of rich text... thoughts?

src/components/RichText/RichText.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Find all anchor elements with target="_blank"
+   *
+   * Regex match groups
+   * (?!.*class="non")               Negative lookahead to filter out any matches which contain this class
+   * (<a[^>]*target="_blank"[^>]*>   Opening anchor tag which has a target="_blank" attribute
+   * ([^<]+)                         The anchor content (any character which is not a left angle bracket)
+   * (<\/a>)                         Closing anchor tag
+   */
```
chriddell: Very helpful, thank you!

src/components/RichText/_text.scss
creido-welly: This file is actually a duplicate of (Text/_text.scss) which I've now removed.

RichText could be used for everything but one thing to bear in mind is that there may be different styles in place for different components

e.g. I can see in your _rte.scss you have bottom margins removed from all RTE child elements which would be at odds with [_text.scss](https://github.com/wellcometrust/corporate-components/blob/develop/src/components/Text/_text.scss)

Happy to standardise though and we can work towards that in future PRs.

chriddell (approved):

PR #285, opened 2020-09-18
7314 bug: FormFieldError logic fixed

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/7314

Fix logic, so only one markup is shown to user


chriddell (approved):
ace - thanks for sorting this

PR #286, opened 2020-09-18
Feat/timeline #7137

creido-welly:


src/components/Timeline/Timeline.stories.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
```
chriddell, line 1: Love the use of Beatles dummy text 👌🏻 👌🏻 👌🏻

src/components/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  return (
+    <div className={classNames}>
+      {title && <h3 className="cc-timeline__title">{title}</h3>}
+      <ol className="cc-timeline__list">
+        {milestones.map((item, index) => (
+          <li
+            // index used to ensure no two <dt><dd> combos have identical keys
+            // eslint-disable-next-line react/no-array-index-key
+            key={`${index}-${item.date}-${item.title}`}
```
chriddell, line 33: A workaround I've found for this eslint flag is to use `index + 1` and use a 1-based index - I'm just as happy with an eslint disable though to be honest, as it can be a frustrating rule!

chriddell (approved):

src/components/Timeline/Timeline.tsx
creido-welly: yeah it's annoying - it exists for good reason it's just a shame it can't read the context the index is in i.e. it'll be unique come what may

PR #288, opened 2020-09-22
Rich text updates

creido-welly:
## In this PR
* Table external spacing
* Text component - added XSS unit test
* Moved generic styles for rich text content elements for greater reusability and to ensure bottom margin for last content element is removed correctly


src/components/RichText/_rich-text.scss
```diff
@@ -60,4 +95,40 @@
...
+    @extend %list-base;
+    margin-bottom: calc(2 * var(--space-unit));
+  }
+
+  > :last-child {
+    margin-bottom: 0;
+  }
+}
+
+.cc-text-snippet__title,
```
chriddell, line 123: is this selector correct? feels wrong/like it shouldn't be here...

chriddell (approved):
left a comment re: cc-text-snippet__title selector

src/components/RichText/_rich-text.scss
creido-welly: I know what you mean but I felt it made less sense to keep _text.scss for a single rule.

I originally attempted to include the [title](https://github.com/wellcometrust/corporate-components/blob/10e7ffa275028b500d0ac67c4c4ab2b940f356b0/src/components/Text/Text.tsx#L25) as a child node of `<RichText>` to keep everything in the same place but was having TypeScript issues

creido-welly: Thinking about it, it now seems like `<Text>` as a component could be redundant. However we still have the title to accommodate which we wouldn't necessarily want to include in the ultra generic `<RichText>`

PR #291, opened 2020-09-23
Check if image exist before adding markup.

DipakShrestha18:


src/components/Contact/Contact.tsx
```diff
@@ -40,15 +40,19 @@ export const Contact = ({
     [`${className}`]: className
   });
 
+  console.log(imageSources);
```
chriddell: Remove the console log (unless this is needed for testing)

chriddell: tbh this shouldn't get past the linter in the pre-commit hook, I'll double-check that and see if we can ensure this doesn't happen again

DipakShrestha18: Don't know why I keep doing it. Removed it.

chriddell (approved):

PR #292, opened 2020-09-23
Tableau title field added to paragraphs mappings

DipakShrestha18:


src/components/TableauChart/TableauChart.tsx
```diff
@@ -52,6 +54,7 @@ export const TableauChart = ({
 
   return (
     <div className={classNames}>
+      <h4>{title}</h4>
```
creido-welly: This should prob be an H4.

If the style needs to be equivalent to H4 this should be handled by adding a class name and a style sheet

DipakShrestha18: Do we have styles for H4?

DipakShrestha18: Can we not use default H4 styling without class?

creido-welly: We do have styles for H4 but these are specific to child elements of .cc-rich-text. By way of explanation, we are aiming to create a component library with selectors which can be overwritten as cleanly as possible, so the CSS must be as flat  as possible. We are also aiming to target HTML elements in style sheets as little as possible for the best flexibility. The only reason we have `.cc-rich-text h4` is to accomodate rich text from a CMS (otherwise we likely wouldn't have it).

BTW you've now done this exactly the way I would have :)

creido-welly (approved):

src/components/TableauChart/TableauChart.tsx
DipakShrestha18: Cheers.

PR #295, opened 2020-09-24
#7339 bug: accordion ui style 

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/7339

Updates accordion's style after QA.

### DIFF
--- a/src/assets/styles/10-settings/_sizes.scss
+++ b/src/assets/styles/10-settings/_sizes.scss
@@ -15,5 +15,5 @@
 
   --secondary-nav-height: calc(2 * var(--space-unit));
 
-  --search-height-md: calc(8 * --var(--space-unit));
+  --search-height-md: calc(8 * var(--space-unit));
 }
--- a/src/components/Accordion/Accordion.stories.tsx
+++ b/src/components/Accordion/Accordion.stories.tsx
@@ -14,6 +14,7 @@ const AccordionExample = () => {
     <>
       <Accordion>
         <AccordionItem
+          key="1"
           title={title}
           titleActive={titleActive}
           variant="chevron"
@@ -25,7 +26,7 @@ const AccordionExample = () => {
       <p>Accordion - the siblings closed when one is open</p>
       <Accordion>
         {[1, 2, 3, 4].map(data => (
-          <AccordionItem title={`${title} ${data}`}>
+          <AccordionItem key={`2-${data}`} title={`${title} ${data}`}>
             {`${content} ${data}`}
           </AccordionItem>
         ))}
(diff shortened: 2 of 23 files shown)


src/components/Accordion/Accordion.stories.tsx
```diff
@@ -33,7 +33,7 @@ const AccordionExample = () => {
       </Accordion>
       <br />
       <p>Accordion - stay open until closed</p>
-      <Accordion>
+      <Accordion className="cc-accordion--0">
```
skibinska: Please check the explanation in this PR: https://github.com/wellcometrust/corporate-react/pull/436

src/components/Accordion/_accordion.scss
```diff
@@ -14,6 +14,20 @@
...
+}
+
+.cc-accordion--0,
+.cc-accordion--only-child {
+  border-top: 1px solid var(--colour-grey-10);
+}
+
+.cc-accordion--last-child,
+.cc-accordion--only-child {
+  margin-bottom: var(--space-lg) !important;
```
skibinska: to overrite `.u-scaffold-grid-item > :not(.cc-card-listing--horizontal):last-child` margin-bottom: 0

skibinska: Now I am thinking to move these styles to corporate-react, as these are accordion modifiers.

creido-welly (approved):

src/components/Accordion/Accordion.tsx
```diff
@@ -6,11 +6,17 @@ import AccordionItem from './AccordionItem';
 type AccordionProps = {
   children?: JSX.Element | JSX.Element[];
   className?: string;
+  inPageSection?: boolean;
```
chriddell, line 9: Added this prop to prevent borders being added to any Accordions used not inside the main page content (e.g. InPageNav).

Previously we had the borders applied through styles on the button inside the Accordion (InPageNav uses the 'chevron' variant, main uses 'plus' variant). This was fine, but because we've moved the borders to the wrapping `cc-accordion` selector* we needed to address this.

*I thought best to move the borders here as I think it's more legible to attach the hack selectors to the top level accordion wrapper, as opposed to a nested button

src/assets/styles/10-settings/_sizes.scss
```diff
@@ -15,5 +15,5 @@
 
   --secondary-nav-height: calc(2 * var(--space-unit));
 
-  --search-height-md: calc(8 * --var(--space-unit));
+  --search-height-md: calc(8 * var(--space-unit));
```
chriddell, line 18: Sorry - this was a bug I found and fixed when I was looking at upgrading postcss-css-variables... felt unnecessary to remove again!

PR #296, opened 2020-09-25
#7364 feat: export css variables

chriddell:
This PR ensures that CSS variables are not compiled to their static values.

See https://github.com/wellcometrust/corporate/issues/7364

- removes postcss-css-variables (and stops build compiling CSS variables to computed values)
- updates rollup
- updates postcss
- updates rollup-postcss-plugin


config/rollup.config.js
```diff
@@ -70,13 +61,9 @@ export default (async () => ({
     }),
     json(),
     postcss({
-      extract: 'dist/style.css',
+      extract: 'style.css',
```
chriddell, line 64: updated version of postcss builds to `dist` folder anyway. (so this was building to `dist/dist/style.css`)

creido-welly (approved):

PR #298, opened 2020-09-28
#7271 feature: report listing page 

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/7271

Show type in `ResultItem` component

Screenshot 2020-09-28 at 13 16 17


src/components/ResultsItem/ResultsItem.tsx
```diff
@@ -49,17 +50,12 @@ export const ResultsItem = ({
     <article className={classNames} id={id}>
       {meta && (meta.type || meta.date) && (
         <div className="result-item__meta">
-          {/* `meta.type` temporarily removed as types for search results are not correctly displayed
-          Type label and date logic for cards has been moved to the app side
-          therefore `itemType` prop has been permanently removed
-          TODO: To be revisited for both search results AND text listings
-          https://github.com/wellcometrust/corporate/issues/6730 */}
-          {/* {meta.type && !['taxonomy_term'].includes(itemType) && (
+          {meta.showType && (
```
chriddell: Are we safe to remove this? Not sure if this issue has been solved... @creido-welly

creido-welly: Not sure. Best to err on the side of caution and leave it in for the time-being

src/components/ResultsItem/ResultsItem.tsx
```diff
@@ -60,6 +61,13 @@ export const ResultsItem = ({
               &nbsp;
             </span>
           )} */}
+          {/* hasType is a flag to decide if we want to show a type */}
```
creido-welly, line 64: Do we actually need this prop? We can either pass in the type or not pass it and then simply have

``` 
{meta.type && (
  ...
)}

skibinska: I only added it because of the comment about types for search results page being not correctly displayed. I didn't want to introduce any bugs to that page. I can remove it if you think we don't need this. :)

creido-welly: Keep it in for now - it shouldn't affect existing results right?

Moving forward we need to get consensus from the product managers on how these items will be labelled and we can always remove if this is handled differently

creido-welly (approved):

PR #313, opened 2020-10-06
#7441 feat: add nav wrapper to pagination component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/7441

pagination updates

### DIFF
--- a/src/components/Grid/GridCell/GridCell.tsx
+++ b/src/components/Grid/GridCell/GridCell.tsx
@@ -3,20 +3,17 @@ import cx from 'classnames';
 
 type GridCellProps = {
   children: React.ReactNode;
-  className?: string;
   column: 1 | 2 | 3;
   columnCount: 1 | 2 | 3;
 };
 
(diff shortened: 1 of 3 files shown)


chriddell (changes requested):
See recent comments on sibling PR https://github.com/wellcometrust/corporate-react/pull/474...

Good idea/decision to add the `<nav>` wrapper to the Pagination, but I don't think we should add a className prop to the `<GridCell>`s, definitely not in this PR at least.

I think good idea to add a className prop to the pagination, though.

chriddell (changes requested):
Sorry for the annoying request... but could you close this PR and create a new one, with the branch name changed to follow the usual convention?

e.g. `feat/add-nav-wrapper-pagination-#7441` || `feat/pagination-accessibility-#7441`

Would suggest checking out a new branch, merging this branch into it, and then creating a PR for the new branch - should save you having to manually copy the changes! 🙂

PR #318, opened 2020-10-09
#7426 feat: decode html entities in ResultItem title

chriddell:
Solves wellcometrust/corporate/issues/7426

- adds html-parser util as wrapper for ReactHtmlParser
- updates ReactHtmlParser usage across app to use new util instead

### DIFF
--- a/src/utils/parse-html.ts
+++ b/src/utils/parse-html.ts
@@ -1,0 +1,23 @@
+/**
+ * @file HTML Parser utility.
+ */
+import 'ts-replace-all';
+import ReactHtmlParser, { Options } from 'react-html-parser';
+
+/**
+ * Parse some HTML using the ReactHtmlParser lib.
+ *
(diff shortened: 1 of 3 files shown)


src/utils/parse-html.ts
```diff
@@ -0,0 +1,25 @@
+/**
```
chriddell, line 1: Reviewers - should this be a component, rather than a util?

creido-welly: I see no reason why it should be a component. It would make it more React-centric but there are no props to handle and no markup to add

creido-welly (approved):

src/utils/parse-html.ts
chriddell: Main point of contention for me was that it returns a ReactElement, so it made me stop and ask the question. I don't think I'm particularly swung to either side, so I'm happy to keep it in the utils.

PR #321, opened 2020-10-13
#7487 feat: adds Author component

chriddell:
Resolves wellcometrust/corporate/issues/7487

- adds Author component
- adds _author.scss styles
- adds Sass placeholder for object-fit/cover images

### DIFF
--- a/src/components/Author/Author.tsx
+++ b/src/components/Author/Author.tsx
@@ -24,23 +24,24 @@ export const Author = ({
     {(imageSrc || imageSrcSet) && (
       <figure className="cc-author__image">
         <ImageElement
+          itemProp="image"
           sizes={imageSizes}
           src={imageSrc}
           srcSet={imageSrcSet}
-          alt={`A photograph of the author: ${name}`}
+          alt={`A photograph of the author, ${name}.`}
         />
       </figure>
     )}
-    <span className="cc-author__body">
+    <div className="cc-author__body">
       <h2 className="cc-author__name" itemProp="name">
         {name}
       </h2>
       <p className="cc-author__byline">
         {jobTitle && <span itemProp="jobTitle">{jobTitle}</span>}
         {jobTitle && organization && ', '}
-        {organization && <span itemProp="affiliation">{organization}</span>}
+        {organization && <span itemProp="memberOf">{organization}</span>}
       </p>
-    </span>
+    </div>
   </div>
 );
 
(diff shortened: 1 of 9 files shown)


src/components/Author/Author.tsx
```diff
@@ -0,0 +1,47 @@
...
+  organization
+}: AuthorProps) => (
+  <div className="cc-author" itemScope itemType="http://schema.org/Person">
+    {(imageSrc || imageSrcSet) && (
+      <figure className="cc-author__image">
+        <ImageElement
+          sizes={imageSizes}
+          src={imageSrc}
+          srcSet={imageSrcSet}
+          alt={`A photograph of the author: ${name}`}
```
creido-welly: Minor: Grammatically a comma would be better than a colon

src/components/Author/Author.tsx
```diff
@@ -0,0 +1,47 @@
...
+        />
+      </figure>
+    )}
+    <span className="cc-author__body">
```
creido-welly: h2 and p shouldn't be wrapped by a span. Suggest using a div instead

src/components/Author/_author.scss
```diff
@@ -0,0 +1,51 @@
...
+.cc-author {
+  margin-bottom: var(--space-lg);
+
+  @include mq($until: md) {
+    display: flex;
+  }
+}
+
+.cc-author__image {
+  --image-size: calc(6.25 * var(--space-unit));
```
creido-welly, line 18: interesting approach 👍  - let me know how you get on with this

src/components/Author/_author.scss
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.cc-author__name,
+.cc-author__byline {
+  @extend %margin-0;
+
+  font-size: var(--body-md);
+}
+
+// .cc-author__body {}
```
creido-welly: Minor: if no styles to add here then remove the extraneous

creido-welly (approved):
All good - couple of comments to look at

src/components/Author/_author.scss
chriddell: works well in our set up I think... because we don't do any CSS variable compiling until corporate-react, in which they're compiled to static values anyway - this means that the variable itself is never exposed outside of the `cc-author__image` selector, so it can only be accessed through that selector (or its descendants)

PR #325, opened 2020-10-15
#7519 feat: PageHeaderCompact component

chriddell:
Resolves wellcometrust/corporate/issues/7519

- adds PageHeaderCompact component

### DIFF
--- a/src/components/PageHeaderCompact/PageHeaderCompact.tsx
+++ b/src/components/PageHeaderCompact/PageHeaderCompact.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import Author, { AuthorProps } from 'Author/Author';
 import { ImageElement } from 'Image';
+import PageTitle from 'PageTitle';
 import RichText from 'RichText';
 import TagList, { TagProps } from 'TagList/TagList';
 import SocialShare from 'SocialShare';
@@ -46,12 +47,7 @@ export const PageHeaderCompact = ({
 }: PageHeaderCompactProps) => (
   <header className="cc-page-header-compact">
     <div className="cc-page-header-compact__main cc-page-header-compact__section cc-page-header-compact__section--main">
-      <span className="cc-page-header-compact__meta">
-        <strong className="cc-page-header-compact__meta-label">
-          {metaLabel}
-        </strong>
-      </span>
-      <h1 className="cc-page-header-compact__title">{title}</h1>
+      <PageTitle metaLabel={metaLabel} title={title} />
       <div className="cc-page-header-compact__standfirst">
         <RichText>{standfirst}</RichText>
       </div>
@@ -61,22 +57,15 @@ export const PageHeaderCompact = ({
         <ImageElement src={imageSrc} srcSet={imageSrcSet} alt={imageAlt} />
         <figcaption className="cc-media__caption">
           {imageCaption && (
-            <div
-              className="cc-media__caption-detail"
-              dangerouslySetInnerHTML={{ __html: imageCaption }}
-            />
+            <RichText className="cc-media__caption-detail">
+              {imageCaption}
+            </RichText>
           )}
           {imageCredit && (
-            <div
-              className="cc-media__credit"
-              dangerouslySetInnerHTML={{ __html: `Credit: ${imageCredit}` }}
-            />
+            <RichText className="cc-media__credit">{imageCredit}</RichText>
           )}
           {imageLicence && (
-            <div
-              className="cc-media__licence"
-              dangerouslySetInnerHTML={{ __html: `Licence: ${imageLicence}` }}
-            />
+            <RichText className="cc-media__license">{imageLicence}</RichText>
           )}
         </figcaption>
       </figure>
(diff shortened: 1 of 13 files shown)


src/components/PageHeaderCompact/PageHeaderCompact.tsx
```diff
@@ -0,0 +1,115 @@
+import React from 'react';
+
+import Author, { AuthorProps } from 'Author/Author';
+import { ImageElement } from 'Image';
+import RichText from 'RichText';
```
creido-welly, line 7: Minor: alphabetical order

src/components/PageHeaderCompact/PageHeaderCompact.tsx
```diff
@@ -0,0 +1,115 @@
...
+import TagList, { TagProps } from 'TagList/TagList';
+import SocialShare from 'SocialShare';
+import FormattedDate from 'FormattedDate';
+
+interface AuthorPropsExtend extends AuthorProps {
```
creido-welly: type is preferable to interface as interface can easily inadvertently acquire props from other components, in which case you can change this to 

```type AuthorPropsExtend = AuthorProps &  {```

Example in use for the Card component
https://github.com/wellcometrust/corporate-components/blob/e4812b145f31b2db9fb84db0460778c6beaddf49/src/components/Card/Card.tsx#L18

File: src/components/PageHeaderCompact/PageHeaderCompact.tsx [hunk 3]
```diff
@@ -0,0 +1,115 @@
...
+  imageLicence?: string;
+  metaLabel?: string;
+  socialUrl: string;
+  standfirst: string;
+  title: string;
+  topics: TagProps[];
+};
+
+export const PageHeaderCompact = ({
+  authors,
```
creido-welly, line 32: Minor: alphabetical order

src/components/PageHeaderCompact/PageHeaderCompact.tsx
chriddell: ah amazing, I thought had to use an interface in order to extend. will update this

src/components/PageHeaderCompact/PageHeaderCompact.tsx
```diff
@@ -0,0 +1,115 @@
...
+  imageCaption,
+  imageCredit,
+  imageLicence,
+  imageSizes,
+  imageSrc,
+  imageSrcSet,
+  metaLabel,
+  title,
+  topics,
+  socialUrl,
```
creido-welly, line 43: Minor: alphabetical order

src/components/PageHeaderCompact/PageHeaderCompact.tsx
```diff
@@ -0,0 +1,115 @@
...
+          {imageLicence && (
+            <div
+              className="cc-media__licence"
+              dangerouslySetInnerHTML={{ __html: `Licence: ${imageLicence}` }}
+            />
+          )}
+        </figcaption>
+      </figure>
+    )}
+    <div className="cc-page-header-compact__aside cc-page-header-compact__section cc-page-header-compact__section--sidebar">
```
creido-welly, line 73: Do you think this could be an `<aside />` element?

chriddell: good question, `<aside>` spec says it...

> represents a portion of a document whose content is only indirectly related to the document's main content

so the question is are authors and topic tags directly or indirectly related to the article content? Imo at the very least, author is directly related

src/components/PageHeaderCompact/PageHeaderCompact.tsx
```diff
@@ -0,0 +1,115 @@
...
+        </div>
+      )}
+      {topics && (
+        <div className="cc-page-header-compact__topics">
+          <TagList tags={topics} />
+        </div>
+      )}
+    </div>
+    <div className="cc-page-header-compact__tray cc-page-header-compact__section cc-page-header-compact__section--main">
+      <time dateTime={date}>
```
creido-welly, line 96: Just looking across this library, I think the `<time>` element could be incorporated into the `<FormattedDate>` component - only ResultsItem does wrap FD with `<time>` and it prob should. Could be a future PR tho.

Can you think of any reason we wouldn't want to wrap it?

chriddell: I have a feeling `<time>` could be the element that causes iOS (and potentially Android/other mobile OS') to allow the user to add the date to their calendar... Not sure if this should be a reason to stop us doing it, though, but could create a style issue?

src/components/PageHeaderCompact/PageHeaderCompact.tsx
creido-welly: True

src/components/PageHeaderCompact/PageHeaderCompact.tsx
creido-welly: Ticket created for this https://github.com/wellcometrust/corporate/issues/7529

We can take a view further down the line whether to do it or not

creido-welly (approved):

PR #326, opened 2020-10-16
#7530 feat: refactors application of img-* extends to apply to child img elements

chriddell:
When working on https://github.com/wellcometrust/corporate/issues/7530 I noticed a discrepancy in how we were using Sass placeholders/extends for generic `img` styles (e.g. %img-responsive, %img-cover). 

All were applied directly to the `` element, however some used a classname selector the hook the styles while others used the child combinator (`> img`) to hook the styles.

This PR regulates the use of those extends by hooking the styles to the `` using the child combinator.

Though usually I would advocate for using a classname to hook these styles to, I feel with ``s there are a couple of advantages to using the `> img` combinator:

- we can apply the styles to "hard-to-reach" `` elements - e.g. from CMS WYSIWYG/RTE
- we don't have to create "wrapper" classname selectors for the `` elements, e.g. `.cc-card__image` & `cc-card__image-wrap`


creido-welly:
@chriddell tnx for the explanation. Standardisation is always a good thing. Can you confirm that image presentation remains good for all affected components in IE11?

> * we can apply the styles to "hard-to-reach"  elements - e.g. from CMS WYSIWYG/RTE

Speaking to @lukewhitmore and @DipakShrestha18, there are currently no plans to allow images within rich text, citing lack of flexibility. Wrapper classnames are a small price to pay and it would certainly be my preference to apply classes to all elements.

chriddell:
No probs - can confirm IE11 still remains good-"ish" (screens below) 😬 ...

As far as where the styles are hooked to... I'm not gonna die on this hill. I think it is _slightly_ easier to remember the selector/markup convention of a single classname wrapping a selector, but I think as long as we stick to the same naming convention with the wrapper (e.g. `block__image-wrap`) then it's much of a muchness.

Live:
Screenshot 2020-10-19 at 14 19 51

This PR:
Screenshot 2020-10-19 at 14 19 49

chriddell:
Also the point about not allowing images in the rich text is a fair one, but is tightly coupled to corporate-react, but our lib should be more broad than the CR usage 🤓

creido-welly:
@chriddell not entirely sure what you mean by tightly coupled to corporate-react

creido-welly (approved):

chriddell:
@creido-welly as in the issue of allowing images through the RTE is a decision for corporate-react, but corporate-components isn't (in theory!) limited to only serve corporate-react... it _could_ be used in other apps in the future, potentially ones which do allow images through the WYSIWYG

PR #331, opened 2020-10-20
#6031 feat: Header nav active links 

creido-welly:
This PR provides a short-term fix to set appropriate header nav links active (both first and second levels) to resolve https://github.com/wellcometrust/corporate/issues/6031. A former corporate codebase script has been repurposed within the library Header component itself to be more self contained and maintainable.

Long-term solution would likely be an in-depth overhaul of the header nav once the Drupal 8 menu APIs are fully featured


chriddell (approved):
👍

PR #334, opened 2020-10-21
7566 feat: updated spacing for nested bullet point

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/7566

Fix spacing issue on the nested bullet point list

Screenshot 2020-10-21 at 10 17 36

Regular list looks ok too:
Screenshot 2020-10-21 at 10 20 42

### DIFF
--- a/src/assets/styles/20-tools/_extends/_lists.scss
+++ b/src/assets/styles/20-tools/_extends/_lists.scss
@@ -35,8 +35,8 @@
 
   ol,
   ul {
-    margin-bottom: 0;
-    margin-top: var(--space-unit);
+    margin-bottom: var(--space-md);
+    margin-top: var(--space-md);
   }
 
   li:not(:last-child) {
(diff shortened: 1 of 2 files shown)


src/assets/styles/20-tools/_extends/_lists.scss
```diff
@@ -35,8 +35,8 @@
 
   ol,
   ul {
-    margin-bottom: 0;
-    margin-top: var(--space-unit);
+    margin-bottom: var(--space-md);
```
creido-welly: Just checking, is it intentional to have the top and bottom margins of the nested lists change over breakpoints? Only asking since the bottom margin of the first level is fixed (line 33)

skibinska: just checked with Jason, and we want them to be 16px at all breakpoints. I will update :)

chriddell (approved):

PR #336, opened 2020-10-21
#7521: Rich article + basic page issues

creido-welly:
Rich article/basic page bugs

#7521 bug: fix to force long links to wrap and prevent horizontal scroll
#7521 feat: Rich text external link icons - updated regex to filter out email addresses and remove unnecessary capturing groups, added external icon specific tests

### DIFF
--- a/src/components/RichText/_rich-text.scss
+++ b/src/components/RichText/_rich-text.scss
@@ -7,8 +7,12 @@
 // ----------------------------------
 
 .cc-rich-text {
+
+  // ensure extra long words are wrapped to prevent overlap or horizontal scroll
+  overflow-wrap: break-word;
   // ensure wide table data can be seen using horizontal scroll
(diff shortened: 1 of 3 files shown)


src/components/RichText/_rich-text.scss
```diff
@@ -7,8 +7,12 @@
 // ----------------------------------
 
 .cc-rich-text {
+
```
chriddell, line 10: Was this space added by the linter? Not sure it should be there, but potentially that's a linting issue 😬

chriddell (approved):

src/components/RichText/_rich-text.scss
creido-welly: yes, it got linted! Not my choice

chriddell: dang linter

PR #337, opened 2020-10-22
#7573 feat: Text anchors (WIP)

creido-welly:
Relates https://github.com/wellcometrust/corporate/issues/7573

- fix to force browser to scroll to position just under the header

@chriddell this is not currently fully tested or perfected but works for most cases with the main exception being IE11. I'm thinking the scroll-to utils in corporate-react could be moved to CC and incorporated here somehow, plus they should still be useable by CR, right?

Tested in Chrome, Firefox, Safari, IE11

### DIFF
--- a/src/components/RichText/_rich-text.scss
+++ b/src/components/RichText/_rich-text.scss
@@ -108,37 +108,6 @@
   > :last-child {
     margin-bottom: 0;
   }
-
-  // Handle anchors
-  //
-  // e.g. <a id="my-anchor"></a>
-  //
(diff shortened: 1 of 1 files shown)


chriddell:
Closed in favour of https://github.com/wellcometrust/corporate-react/pull/536

@creido-welly I felt that this work belonged in corporate-react due it being related to the wellcome.org codebase/CMS specifically, and didn't feel right that the "hack" we're implementing should be in a component library which may be used in other projects.

PR #346, opened 2020-10-28
7645 feat: add contactForm prop to InfoBox

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/7645

The current Contact us has an option to send a message. It is a form not an email

Screenshot 2020-10-28 at 13 03 53

link: https://wellcome.org/grant-funding/guidance-listing

I have decided to implement this by adding `contactForm` to `InfoBox` component

Screenshot 2020-10-28 at 13 14 05

Please let me know if this is the correct way

### DIFF
--- a/src/components/Contact/Contact.tsx
+++ b/src/components/Contact/Contact.tsx
@@ -15,7 +15,7 @@ type ContactProps = {
   className?: string;
   contactFormLink?: {
     text?: string;
-    url?: string;
+    url: string;
   };
   email?: string;
   imageSources?: ContactImageSourceProps;
--- a/src/components/Contact/Contact.tsx
+++ b/src/components/Contact/Contact.tsx
@@ -22,7 +22,7 @@ type ContactProps = {
   institution?: string;
   institutionCountry?: string;
   contactRole?: string;
-  name: string;
+  name?: string;
   teamTitle?: string;
   teamUrl?: string;
   tel?: string;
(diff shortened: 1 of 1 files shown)


skibinska:
@chriddell, what do you think about that?

chriddell (commented):
Makes sense to me.

~One thing I would ask - is it used in multiple places, with different titles/urls? As far as I can see it only always links to the same place, with the same text ("Send a message").~

~If so, I think a boolean prop would be more appropriate e.g. `isLinkToContactForm`~

Ignore me! Realised this is in corporate-components, so we should allow for flexibility!

chriddell (commented):
Looks good - I have some reservations about the name `contactForm` - could we call it `contactFormLink` instead, to make it clear that it's a link to the contact form, rather than rendering the actual form inside the component?

It would be helpful to also declare a default for the 'title' (rename to 'text', as 'title' makes it sound like it is rendered inside a `<h>` tag). 

e.g.

```jsx
<a href={`${contactFormLink.url}`} className="cc-contact__link">
  contactFormLink.text || 'Send a message'
</a>
```

src/components/Contact/Contact.tsx
```diff
@@ -13,19 +13,24 @@ type ContactImageSourceProps = {
 
 type ContactProps = {
   className?: string;
+  contactFormLink?: {
+    text?: string;
+    url?: string;
```
chriddell: url should be required (I think)

chriddell (approved):
👍

PR #351, opened 2020-11-04
#7563 feat: updates storybook 6.0.28

chriddell:
See wellcometrust/corporate/issues/7563

- updates `storybook` and `@storybook/` packages to 6.0.28 (latest stable)
- adds `html-webpack-plugin@5.0.0-alpha.9` to dev dependencies (this is a workaround for https://github.com/jantimon/html-webpack-plugin/issues/1377 & https://github.com/storybookjs/storybook/issues/9216)
- updates storybook config implementation to match best practice from Storybook 6.0.x docs


package.json
```diff
@@ -64,6 +64,7 @@
     "react-html-parser": "^2.0.2",
     "react-paginate": "^6.5.0",
     "shortid": "^2.2.15",
+    "storybook": "^6.0.28",
```
creido-welly: should this be under `devDependencies`?

creido-welly (approved):
All good, just that one query over dependency vs devDependency

package.json
chriddell: Good q, though I think we deploy to https://ghp.wellcomecollection.org/corporate-components/ - so does that mean it should stay in depdendencies?

chriddell: ...then again, that URL is not really production, is it? It's just a deployed version of a testing environment. I think I'll move storybook to devDependencies.

creido-welly: Absolutely. The storybook dependency would become part of the npm package which we definitely don't want

creido-welly (approved):

PR #352, opened 2020-11-06
#7671 feat: ImageCard component (NewsCard)

chriddell:
See wellcometrust/corporate/issues/7671

- adds NewsCard component
- adds NewsCard storybook component
- adds `width` and `height` prop to ImageElement (handy prep work for Next JS Image component https://github.com/wellcometrust/corporate/issues/7682)
- adds `%anchor-reverse` extend
- adds `height: auto;` to `%img-responsive` extend
- adds `--colour-amber-30` to `_colours.scss` settings


src/components/NewsCard/NewsCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+  image: {
+    alt: string;
+    height?: string;
+    sizes?: string;
+    src: string;
+    srcSet?: string;
+    width?: string;
+  };
+  title: string;
+  titleAs?: keyof JSX.IntrinsicElements;
```
creido-welly: :+1: like this idea

It would be good to limit this prop to headings only (and probably leave out h1)

src/components/NewsCard/NewsCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { ImageElement } from 'Image';
+import FormattedDate from 'FormattedDate';
+import Link from 'Link';
+
+type NewsCardProps = {
+  author: string;
+  className?: string;
+  date: string;
+  image: {
+    alt: string;
```
creido-welly: I'm becoming less of a fan of these nested props as it makes them less visible. PageHeaderCompact has a very flat structure and it's instantly more readable.

I also think over time it would be good to refactor other components which have nested props.

chriddell: good point - I think you're right, have updated

src/components/NewsCard/NewsCard.tsx
chriddell: not sure this it's possible to grab just the headings keys from this, so have just declared more explicitly what is allowed!

agreed would be nice to be able to get a subset of IntrinsicElements, but think this does the job just as well anyway - a good one to know for the future

creido-welly (approved):

src/assets/styles/20-tools/_mixins/_typography.scss
```diff
@@ -27,25 +27,55 @@
...
+      transition: all var(--transition-duration) var(--transition-timing-function-in);
+      visibility: hidden;
+      width: 100%;
+    }
+
+    &:hover:before {
+      transform: scaleX(1);
+      visibility: visible;
+    }
+  }
```
chriddell, line 65: Want to check in with @j-jaworski on this one... we're introducing a new, v similar style with the animated underline (imitated from British Museum website) to one we currently use in the footer nav links.

Will check in on Monday to see if we want to replace the footer nav link style with the new updated one, or whether we need to maintain both. I'll then update code accordingly.

j-jaworski: We should change the footer hover state to animate from the left as per the new style but think it will look weird if they are exactly the same (thicker line used on card hovers might look a bit OTT on the footer's smaller text links)

chriddell: Ah amazing, wasn't expecting you to respond on here - was gonna flag on Slack. Thanks for that update, v helpful - will keep the line thickness the same as it currently is on footer links (think it's probs 1px).

j-jaworski: 1px sounds right, feel free to make everything else the same, except colour obvs

chriddell: https://github.com/wellcometrust/corporate/issues/7729

PR #355, opened 2020-11-12
bug: Storybook compatibility with npm link

creido-welly:
Upgraded storybook to release candidate for v6.1 to fix compatibility issue with "npm link".

Explanation: "npm link" is used heavily in development to link corporate-components to corporate-react as a live reloading dependency


chriddell (commented):
Noticed we've lost ~40,000 lines from package-lock... have you pushed this without re-running `npm i` after npm linking corporate-components/react?

creido-welly:
I'm pretty sure I re-ran npm i but will do it again just to be safe

chriddell:
Could just be that storybook has way less dependencies...

chriddell (approved):

creido-welly:
Slight difference but there's still a discrepancy of over 40,000 lines

PR #358, opened 2020-11-18
#7672 feat: full width promo 

creido-welly:
Relates to https://github.com/wellcometrust/corporate/issues/7672

- Adds new full width promo component as featured on news landing page
- Adds storybook

@chriddell the CTA is currently an integral part of this component but should likely be abstracted as a reusable component since it'll feature elsewhere in future.


chriddell (commented):
The package-lock has lost ~50,000 lines... is ~corporate-components~ React still symlinked?

src/assets/styles/10-settings/_transition.scss
```diff
@@ -9,4 +9,6 @@
 :root {
   --transition-duration-fast: 0.2s;
   --transition-duration: 0.4s;
+  --transition-timing-function-in: cubic-bezier(0.23, 1, 0.32, 1);
+  --transition-timing-function-out: cubic-bezier(0.55, 0.085, 0.68, 0.53);
```
chriddell: Not sure about setting this at the root... I know it needs to be shared between components, but setting at root seems to indicate it's a global variable, whereas it currently is only used in the Card components (and I suspect we have other transition timing function values declare elsewhere.

Any thoughts?

src/components/FullWidthPromo/FullWidthPromo.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { storiesOf } from '@storybook/react';
+import { boolean, select, text } from '@storybook/addon-knobs';
+
+import FullWidthPromo from './FullWidthPromo';
+
+const FullWidthPromoExample = () => {
+  const generalID = 'General';
+  const imageID = 'Image';
+  const metaID = 'Meta';
+
```
chriddell, line 11: I like this idea. Think we should adopt on all stories going forwards

src/components/FullWidthPromo/FullWidthPromo.tsx
```diff
@@ -0,0 +1,92 @@
...
+  imageSrcSet,
+  linkText,
+  metaLabel,
+  title,
+  topics
+}: FullWidthPromoProps) => (
+  <div className="cc-fw-promo">
+    {(!!imageSrc || (imageSrcSet && imageSizes)) && (
+      <figure className="cc-fw-promo__image">
+        <ImageElement src={imageSrc} srcSet={imageSrcSet} alt={imageAlt} />
```
chriddell: Do you think we should provide `height` and `width` to the ImageElement?

Should help if/when we migrate to the NextJS Image component. Regardless, it would also help with lazy-loading if/when we implement that, too.

src/components/FullWidthPromo/FullWidthPromo.tsx
```diff
@@ -0,0 +1,92 @@
...
+  imageAlt,
+  imageSizes,
+  imageSrc,
+  imageSrcSet,
+  linkText,
+  metaLabel,
+  title,
+  topics
+}: FullWidthPromoProps) => (
+  <div className="cc-fw-promo">
```
chriddell: What's the benefit of abbreviating the classname? 

I'm thinking that it would make it more difficult to track down the React component if, say, you were only looking at the rendered HTML and wanted to find the relevant component in the codebase.

src/components/FullWidthPromo/FullWidthPromo.tsx
```diff
@@ -0,0 +1,92 @@
...
+          {linkText}
+        </Button>
+        {topics && (
+          <div className="cc-page-header-compact__topics">
+            <TagList tags={topics} />
+          </div>
+        )}
+      </div>
+    </div>
+  </div>
```
chriddell: What do you think about adding an `itemScope` (and some `itemProp`) attribute(s) to some of the DOM elements?

It seems like the component will always be used for articles, so maybe we can even render it using an `<article>` element?

I think, even if we don't tie the component to an article type, we could use the itemType of `Thing` https://schema.org/Thing

src/components/FullWidthPromo/_full-width-promo.scss
```diff
@@ -0,0 +1,165 @@
...
+  background: var(--colour-amber-05);
+  position: relative;
+}
+
+.cc-fw-promo__image {
+  @extend %img-cover;
+  margin: 0;
+  
+  @include mq($until: sm) {
+    height: calc(220vw / 320 * 100);
```
chriddell: Could you write a comment in the code explaining this calc? I can't work out what it's doing, or why 😬

src/components/FullWidthPromo/_full-width-promo.scss
```diff
@@ -0,0 +1,165 @@
...
+    position: absolute;
+    right: 0;
+    top: 0;
+    z-index: 0;
+  }
+}
+
+.cc-fw-promo__container {
+  @include grid-container;
+  padding: calc(7.5 * var(--space-unit)) 0;
```
chriddell: I think the shorthand may be detrimental here? Seems like we only want to set top/bottom whereas this explicitly sets left/right to 0 as well, which:

- adds unnecessary style declaration
- makes it harder to overwrite (depending on another selectors specificity/order)

src/components/FullWidthPromo/_full-width-promo.scss
```diff
@@ -0,0 +1,165 @@
...
+
+  @include mq(sm) {
+    padding: calc(9 * var(--space-unit)) 0;
```
chriddell: Same as https://github.com/wellcometrust/corporate-components/pull/358/files#r526036302

src/components/FullWidthPromo/_full-width-promo.scss
```diff
@@ -0,0 +1,165 @@
...
+  @extend %anchor-inherit;
+  @include animated-underline;
+}
+
+.cc-fw-promo__meta {
+  align-items: center;
+  display: flex;
+  flex-wrap: wrap;
+  margin-bottom: calc(2 * var(--space-unit));
+  padding: var(--space-unit) 0;
```
chriddell: Same as https://github.com/wellcometrust/corporate-components/pull/358/files#r526036302

src/components/FullWidthPromo/_full-width-promo.scss
```diff
@@ -0,0 +1,165 @@
...
+  font-size: var(--body-xs);
+  margin: 0;
+  text-transform: uppercase;
+}
+
+.cc-fw-promo__type + .cc-fw-promo__authors:before {
+  color: var(--colour-amber-30);
+  content: '/';
+  font-size: var(--body-xs);
+  padding: 0 var(--space-unit);
```
chriddell: Similar to https://github.com/wellcometrust/corporate-components/pull/358/files#r526036302

src/assets/styles/10-settings/_transition.scss
creido-welly: I had similar thoughts as you initially. It makes sense to scope custom props locally where they're clearly meant for for specific components but these underline properties are used in footer nav links too.

I've moved these vars to the parent full width promo selector now but we perhaps need to think about a strategy for future (beyond the scope of this ticket). How about ...

- For the most part other timing functions are simply 'ease'. This could be added as `--transition-timing-default: ease`
- 'ease-in-out' is used for things related to the header, but not just in the header - could be globally scoped as `--transition-timing-header`
- The above could be renamed more specifically `--transition-timing-underline-in` and `--transition-timing-underline-out` and added to the global scope 
- Gallery lightbox and logo use specific cubic-beziers so could use locally scoped properties

src/components/FullWidthPromo/FullWidthPromo.stories.tsx
creido-welly: yes where it makes sense to i.e. loads of props

src/components/FullWidthPromo/FullWidthPromo.tsx
creido-welly: Simply for brevity. TBH I'm not entirely sold on the naming of this component since 'Full Width' is related to style. Thoughts on that? PromoCard? Or is FullWidthPromo fine?

Once we've settled on the name I'll make these easier to read

src/components/FullWidthPromo/FullWidthPromo.tsx
creido-welly: Done

src/assets/styles/10-settings/_transition.scss
chriddell: I like that idea, I think using a specific settings file to hold global transitions makes it clear for codebase novices where they should declare any re-used transition styles.

I wonder if locally scoped variables would benefit from being namespaced, e.g. `--component-transition-timing`.

src/components/FullWidthPromo/FullWidthPromo.tsx
creido-welly: Done

src/components/FullWidthPromo/_full-width-promo.scss
creido-welly: Removed. I added it initially to force the image to display in a fixed aspect ratio but it's probably not that important

src/components/FullWidthPromo/FullWidthPromo.tsx
chriddell: Ahh that's a good point. I totally agree. 

I'm reluctant to name it anything but FullWidthPromo without discussing with design. Potential names (all a bit rubbish):

- SpecialPromo
- ImagePromo
- BigPromo

cc: @j-jaworski

src/components/FullWidthPromo/FullWidthPromo.stories.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
+import { storiesOf } from '@storybook/react';
+import { boolean, select, text } from '@storybook/addon-knobs';
```
chriddell: Looks like `select` and `boolean` aren't used

src/components/FullWidthPromo/FullWidthPromo.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+
+import { ImageElement } from 'Image';
+import Button from 'Button';
+import Icon from 'Icon';
```
chriddell: `Icon` is not used

src/assets/styles/10-settings/_transition.scss
chriddell: Shall I create a ticket for this?

creido-welly: > Shall I create a ticket for this?

Yes please

creido-welly: > I wonder if locally scoped variables would benefit from being namespaced, e.g. `--component-transition-timing`.

I'm open to this in principle however I think it depends on the context. Definitely we should do it where it makes sense to but it's worth reading the following article to see if there's another way which might complement our codebase more. 

https://css-tricks.com/breaking-css-custom-properties-out-of-root-might-be-a-good-idea/

chriddell: ace, will read

chriddell (commented):
Final check that the package-lock is okay? It's lost a lot of lines...

creido-welly:
> Final check that the package-lock is okay? It's lost a lot of lines...

I'm investigating ...

chriddell (commented):
Preliminary approval as I think the only issue left is package-lock, and you're already looking in to it

chriddell (approved):

creido-welly:
> Preliminary approval as I think the only issue left is package-lock, and you're already looking in to it

@chriddell node_modules and package-lock.json deleted and `npm i` run several times using node v10.15.0. I've fixed the react and react-dom versions at v16.12.0 (for some reason my install was defaulting to v16.14.0) and package-lock is still coming out around 50,000 lines shorter

chriddell:
ah okay, I guess something has shedded a lot of dependencies then - all good!

PR #359, opened 2020-11-18
#7742 feat: authors prop is now array of strings (ImageCard)

chriddell:
See wellcometrust/corporate/issues/7742

- updates ImageCard's authors prop to become array of strings (rather than string)
- updates relevant markup
- updates relevant styles

### DIFF
--- a/src/components/ImageCard/ImageCard.tsx
+++ b/src/components/ImageCard/ImageCard.tsx
@@ -6,7 +6,7 @@ import FormattedDate from 'FormattedDate';
 import Link from 'Link';
 
 type ImageCardProps = {
-  author: string;
+  authors: string[];
   className?: string;
   date: string;
   imageAlt: string;
@@ -22,7 +22,7 @@ type ImageCardProps = {
 };
 
 export const ImageCard = ({
-  author,
+  authors,
   className,
   date,
   imageAlt,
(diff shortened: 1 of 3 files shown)


creido-welly (approved):

src/components/ImageCard/ImageCard.tsx
```diff
@@ -62,15 +62,23 @@ export const ImageCard = ({
...
           </span>
-          <span
-            className="cc-image-card__meta-item cc-image-card__meta-item--author"
-            itemProp="author"
-          >
-            {author}
-          </span>
+          {authors && (
+            <dl className="cc-image-card__meta-item cc-image-card__meta-item--author cc-image-card__authors">
+              <dt className="cc-image-card__authors-label">Author</dt>
```
skibinska, line 70: I am wondering if it should change to `Authors` when more than one author. (I haven’t seen the design for this, so please ignore if this is not relevant)

chriddell: Ah it's interesting you raised that... it's not relevant in the way you think it might be...

The designs for these are simply an author's name (e.g. below), but it's interesting that you thought of the Author component, and makes me think maybe this prop would be better declared as `authorNames`...

Screenshot 2020-11-18 at 16 33 53

chriddell: I have a job to consolidate some of the code for the new Card components, and the author names are potentially going to be extracted a new component so I will bare this in mind!

PR #362, opened 2020-11-18
#7749 feat: adds ImageCardMultiple storybook story

chriddell:
This PR adds a storybook component to render multiple ImageCard components so that we can get them QA'd by design.

We do not yet have the style rules written for this, so I have used css-in-js to apply what are temporary styles which will be replicated in Sass once we implement https://github.com/wellcometrust/corporate/issues/7750

See wellcometrust/corporate/issues/7749

- adds storybook story to display multiple ImageCard components

### DIFF
--- a/src/components/ImageCard/ImageCard.stories.tsx
+++ b/src/components/ImageCard/ImageCard.stories.tsx
@@ -69,6 +69,15 @@ const MultipleImageCard = () => {
   return (
     <div
       style={{
+        /**
+         * todo: Render ImageCard with Listing
+         *
+         * This will mean we can render multiple ImageCard components
+         * by passing the data to the Listing component, and move the styles
+         * + DOM from here to be handled by that component + Sass (for the styles)
+         *
+         * @see {@link https://github.com/wellcometrust/corporate/issues/7750}
+         */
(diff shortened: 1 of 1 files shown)


src/components/ImageCard/ImageCard.stories.tsx
```diff
@@ -39,6 +39,71 @@ const ImageCardExample = () => {
...
+    url: text('URL', '/news/all', `Card ${i + 1}`)
+  }));
+
+  return (
+    <div
+      style={{
+        display: 'flex',
+        justifyContent: 'space-between',
+        flexWrap: 'nowrap',
+        marginLeft: 'calc(-1 * var(--space-xl))'
```
chriddell, line 84: I've added these styles as a temporary solution for displaying multiple ImageCard components in a row so that we can get these components QA'd by design.

I feel like this is appropriate for now because it prevents having to write the style rules elsewhere in the app - which means it will be easier to tidy up later.

creido-welly (approved):

chriddell:
Closing as implemented in a more appropriate manner in https://github.com/wellcometrust/corporate-components/pull/362

PR #365, opened 2020-11-25
#7773 feat: aligns prop names/types in FeaturedPromo and FullWidthPromo

chriddell:
Aligns prop names + types across FeaturedPromo and FullWidthPromo components

See wellcometrust/corporate/issues/7773

- updates various prop names + types in FeaturedPromo
- updates DOM in FeaturedPromo (to allow for updated `authors` prop - is now an array)
- updates styles in FeaturedPromo (to align with above DOM)
- updates `imageSrcSet` prop to optional in FullWidthPromo
- updates `imageSrc` prop to required in FullWidthPromo
- updates FullWidthPromo storybook component

### DIFF
--- a/src/components/FeaturedPromo/FeaturedPromo.tsx
+++ b/src/components/FeaturedPromo/FeaturedPromo.tsx
@@ -9,6 +9,7 @@ type FeaturedPromoProps = {
   authors?: string[];
   className?: string;
   description?: string;
+  href: string;
   imageAlt: string;
   imageHeight?: string;
   imageSizes?: string;
@@ -18,7 +19,6 @@ type FeaturedPromoProps = {
   metaLabel?: string;
   title: string;
   titleAs?: 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
-  href: string;
 };
 
 export const FeaturedPromo = ({
--- a/src/components/FeaturedPromo/FeaturedPromo.tsx
+++ b/src/components/FeaturedPromo/FeaturedPromo.tsx
@@ -6,7 +6,7 @@ import RichText from 'RichText';
 import Link from 'Link';
 
 type FeaturedPromoProps = {
-  author: string;
+  authors?: string[];
   className?: string;
   description?: string;
   imageAlt: string;
@@ -15,26 +15,25 @@ type FeaturedPromoProps = {
   imageSrc: string;
   imageSrcSet?: string;
   imageWidth?: string;
+  metaLabel?: string;
   title: string;
   titleAs?: 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
-  topic: string;
-  url: string;
+  href: string;
 };
 
 export const FeaturedPromo = ({
-  author,
+  authors,
   className,
   description,
+  href,
   imageAlt,
   imageHeight,
-  imageSizes,
   imageSrc,
   imageSrcSet,
   imageWidth,
+  metaLabel,
   title,
-  titleAs = 'h3',
-  topic,
-  url
+  titleAs = 'h3'
 }: FeaturedPromoProps) => {
   const TitleElement = titleAs;
   const classNames = cx('cc-featured-promo', {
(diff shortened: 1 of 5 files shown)


src/components/FeaturedPromo/FeaturedPromo.tsx
```diff
@@ -15,26 +15,25 @@ type FeaturedPromoProps = {
   imageSrc: string;
   imageSrcSet?: string;
   imageWidth?: string;
+  metaLabel?: string;
   title: string;
   titleAs?: 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
-  topic: string;
-  url: string;
+  href: string;
```
creido-welly: minor: alphabetical order

src/components/FeaturedPromo/FeaturedPromo.tsx
```diff
@@ -47,13 +46,12 @@ export const FeaturedPromo = ({
...
       itemType="https://schema.org/Article"
     >
-      <Link className="cc-featured-promo__figure" to={url}>
+      <Link className="cc-featured-promo__figure" to={href}>
         <figure className="cc-featured-promo__image">
           <ImageElement
             alt={imageAlt}
             height={imageHeight}
             itemProp="image"
-            sizes={imageSizes}
```
creido-welly, line 56: `sizes` is needed to make srcSet work correctly right?

src/components/FullWidthPromo/FullWidthPromo.tsx
```diff
@@ -15,8 +15,7 @@ type FullWidthPromoProps = {
   imageCredit?: string;
   imageHeight?: string;
   imageLicence?: string;
-  imageSizes?: string;
```
creido-welly, line 18: Same comment as above re `sizes` attribute, shouldn't it be kept for srcSet usage?

creido-welly (commented):
Query about removal of sizes attribute

Maybe this helps to inform that decision?
https://css-tricks.com/sometimes-sizes-is-quite-important/

src/components/FeaturedPromo/FeaturedPromo.tsx
chriddell: Yep, but I think it should be set within the component (as the component styles define the width of the image...). Does that make sense?

I just need to add a ticket for defining these sizes (which I'll do if you agree with above)

creido-welly: From that point of view it makes total sense to define it within the component 👍 . Just need to consider, are there any contexts where they might present differently? (none I can think of)

I was torn about whether or not to have sizes defined in the library or have them passed thru from the consuming app, but makes sense that it's a part of its styling and gets defined here. This is how it's been done for [Card.tsx](https://github.com/wellcometrust/corporate-components/blob/develop/src/components/Card/Card.tsx).

Further thought, would it be useful to perhaps have a config file listing all of the sizes together ?!?!?!?

e.g.

```javascript
// /src/constants/img-sizes.ts

const IMG_SIZES = {
  card: '(min-width: 1494px) 420px, (min-width: 1024px) 27vw, (min-width: 768px) 37vw, 90vw',
  imageCard: '...',
  galleryThumbnail: '...',
  ...
}

export default SIZES
```

chriddell: I don't think there are any contexts where we allow them to present differently (at least not currently).

Not sure about abstracting the config... I'm not sure where it would be used other than within relevant components (ie. Card.tsx only uses `card` key of `IMG_SIZES`, ImageCard.tsx only uses `imageCard` key of `IMG_SIZES` etc.). 

Seems like the abstraction would cause more confusion than declaring the values within each component file?

src/components/FeaturedPromo/FeaturedPromo.tsx
chriddell: man we need a linting rule for this

src/components/FeaturedPromo/FeaturedPromo.tsx
creido-welly: I was spitballing more than anything - sizes can be a pain to wrap your head around at times so this would really be a means of comparison between component types. No point having config for config's sake

creido-welly (approved):

PR #369, opened 2020-11-30
#7795 feat: bg color featured promo image

chriddell:
See wellcometrust/corporate/issues/7795

- adds bg-color to all `%img-*` extends
- applies height + bg-color to featured promo image

### DIFF
--- a/src/assets/styles/20-tools/_extends/_images.scss
+++ b/src/assets/styles/20-tools/_extends/_images.scss
@@ -12,6 +12,7 @@
  */
 %img-responsive {
   > img {
+    background-color: var(--colour-amber-05);
     display: block;
     height: auto;
     width: 100%;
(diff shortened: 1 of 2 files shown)


src/assets/styles/20-tools/_extends/_images.scss
```diff
@@ -12,6 +12,7 @@
  */
 %img-responsive {
   > img {
+    background-color: var(--colour-amber-05);
```
chriddell, line 15: This adds a background color to wherever this extend is used (Card, Author, ImageCard, FeaturedPromo).

I think this is a useful pattern as it means wherever an image fails to load we are consistent in how it presents - meaning the user will start to understand what is going on.

creido-welly (approved):

PR #376, opened 2020-12-02
#7807 feat: refactor ResultsItem as TextCard

creido-welly:
Relates to [#7807](https://github.com/wellcometrust/corporate/issues/7807)

- Renaming of component to bring it in line with recent UX/UI work on cards
- Refactoring of props for flatter, more readable structure in line with recent addition of ImageCard
- Applied changes to ListingLink (link list content variant)


src/components/TextCard/TextCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+              {metaLabel}
+              &nbsp;
+            </span>
+          )} */}
+          {/* hasMetaLabel is a flag to decide if we want to show a type */}
+          {hasMetaLabel && (
+            <span className="cc-text-card__meta-label">
+              {metaLabel}
+              &nbsp;
+            </span>
```
chriddell: Is the boolean prop necessary? Can we set the condition that if `metaLabel` is not falsy then we display it?

creido-welly: It is used in a couple of places so I've erred on the side of caution and kept it in

creido-welly: I don't think it's quite as simple as if metaLabel is present or not as there are clearly instances where it is meant to be displayed and others that aren't

chriddell (approved):

PR #377, opened 2020-12-02
#7652 feat: event action downloads

chriddell:
In order for analytics events to fire on file download links in the same manner as on D7, we need to add some `data-*` attributes to the `<a>` tag. Allows corporate-react to pass props a la https://github.com/wellcometrust/corporate-react/pull/620.

See wellcometrust/corporate/issues/7652

- adds `documentType`, `documentSubType` props to:
  - `<ListingLink />`
  - `<ResultItem />`
  - `<FileDownload />`
- delcares spread props on `Link` component's type

### DIFF
--- a/src/components/FileDownload/FileDownload.tsx
+++ b/src/components/FileDownload/FileDownload.tsx
@@ -3,15 +3,19 @@ import cx from 'classnames';
 
 type FileDownloadProps = {
   className?: string;
+  gaFileType?: string;
+  gaReportType?: string;
   href: string;
   label?: string;
   name?: string;
   size?: string;
-  type?: string;
+  type: string;
 };
 
 export const FileDownload = ({
   className,
+  gaFileType,
+  gaReportType,
   href,
   label = 'Download',
   name,
@@ -26,6 +30,9 @@ export const FileDownload = ({
     <div className={classNames}>
       <a
         className="cc-file-download__link"
+        // todo: #7814
+        data-file-type={gaFileType}
+        data-report-type={gaReportType}
         download
         href={href}
         rel="noopener noreferrer"
--- a/src/components/FileDownload/FileDownload.tsx
+++ b/src/components/FileDownload/FileDownload.tsx
@@ -3,8 +3,8 @@ import cx from 'classnames';
 
 type FileDownloadProps = {
   className?: string;
-  gaFileType?: string;
-  gaReportType?: string;
+  documentSubType?: string;
+  documentType?: string;
   href: string;
   label?: string;
   name?: string;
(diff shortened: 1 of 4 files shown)


src/components/FileDownload/FileDownload.tsx
```diff
@@ -3,15 +3,19 @@ import cx from 'classnames';
...
 type FileDownloadProps = {
   className?: string;
+  documentSubType?: string;
+  documentType?: string;
   href: string;
   label?: string;
   name?: string;
   size?: string;
-  type?: string;
+  type: string;
```
stuartwilkes, line 12: This doesn't cause any unintended 500 responses does it?

stuartwilkes (approved):
Assuming we're good with the change I added a comment to - all approved...

src/components/FileDownload/FileDownload.tsx
chriddell: I don't think so... this is just type-checking and isn't run on production anyway. 

I noticed that we're rendering the `type`  (PDF, XLS etc.) without a conditional, so updated the type declaration for it to be required

src/components/FileDownload/FileDownload.tsx
```diff
@@ -26,6 +30,9 @@ export const FileDownload = ({
     <div className={classNames}>
       <a
         className="cc-file-download__link"
+        // todo: #7814
+        data-file-type={documentType}
```
creido-welly, line 34: Is it worth namespacing data attributes which are concerned with analytics so they are easily distinguishable from non analytics ones? e.g. `data-analytics-file-type` or `data-tracking-file-type`

chriddell: Yep, this is exactly what I want to do... but would also require changes to:

- GTM
- D7

I've deliberately not applied namespacing to save myself the work, because I think we'd also need to undertake some investigation in to where else these specific data attributes are used...

A thought though: I could add both (namespaced and not namespaced) and create a ticket to remove? Or we could just pick up [this ticket](https://github.com/wellcometrust/corporate/issues/7814) when the migration is complete and we're not straddling two codebases?

chriddell: *also realise that ticket needs fleshing out!

creido-welly: Keep the code lean for now and add these data attribute names (and any others you come across) to the opening comment of [#7814](https://github.com/wellcometrust/corporate/issues/7814) so we have a record

creido-welly (approved):

PR #380, opened 2020-12-08
#7793 bug(style): refactors Media component

chriddell:
See wellcometrust/corporate/issues/7793

- refactors Media component to use Grid, GridCell
- removes layout/grid styles from _media.scss

### DIFF
--- a/src/components/Image/Image.tsx
+++ b/src/components/Image/Image.tsx
@@ -32,7 +32,6 @@ export const Image = ({
     className={className}
     credit={credit}
     licence={licence}
-    type="image"
   >
     <ImageElement alt={alt} sizes={srcSet && sizes} src={src} srcSet={srcSet} />
   </Media>
--- a/src/components/Media/Media.tsx
+++ b/src/components/Media/Media.tsx
@@ -1,7 +1,8 @@
-import React, { Children, cloneElement } from 'react';
+import React from 'react';
 import cx from 'classnames';
 
 import MediaCaption from './MediaCaption';
+import Grid, { GridCell } from '../Grid';
 
 type MediaProps = {
   caption?: string;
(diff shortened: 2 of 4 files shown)


creido-welly:
Just out of interest why are you making this change?

chriddell:
In a previous commit I removed the className prop from the ImageElement component, which meant the `` inside the Media component was not being set on the grid (see https://master.wellcome-corporate-fe.org.uk/news/what-are-monoclonal-antibodies-can-they-treat-covid-19).

I think maintaining a rule of not applying classnames to `` elements is sensible as in my experience they can go awry - e.g. style rules not obeyed/causing unexpected behaviour. 

Annoyingly I can't think of specific scenarios, but it has become a habit which I'm devoted to - but maybe I need to re-assess my devotion...

Anyway - I saw that the Media component set up it's own grid in it's own sass file - I thought it would be clearer to use the Grid component within the Media component as it is based on the same grid layout* as the main grid, anyway.

*except that it is in inset by 1 column between `sm` and `md` breakpoints, hence adding https://github.com/wellcometrust/corporate-components/pull/380/files#diff-2312b5c88df635e1d9930e14bc3588b39422c1c061222fa5a324cbbfcda6f54dR11-R16

creido-welly:
Media was originally set up this way so that the grid was an abstraction which could be overridden by the consuming app (using CSS) if circumstances dictate it needs something different from how it's currently set in the design system. Once we set layout using JSX components it's effectively locked in. Same thinking was applied to the image gallery thumbnail.

chriddell:
Yeah that makes sense - I've no preference either way as to where the layout lives (either in component markup/Grid, or in _media styles). I'll put the styles back in to _media.scss, with according tweaks made to ensure the images still live on the grid.

creido-welly:
Cool, tnx. I can completely see why you went down that route.

We ideally need these media and image gallery components to be inherently composable within the consuming app - it's not there yet of course, but moving forward perhaps there is a more clever but simpler solution to the whole scaffolding / grid layout issue.

creido-welly (approved):

PR #381, opened 2020-12-11
#7780 feat: implement responsive Iframe component

aliceh75:
The Iframe component is made responsive with a fairly well documented technique (add a wrapper div, add padding-top to set aspect ratio, and make iframe absolutly positioned within it). Tested with Google Maps (only current use case) on FF and Chrome on Mac. Not tested with Edge yet.

Note that while the component is provided with width/height, it does not enforce the values - it uses them to calculate the aspect ratio. I felt it was best to leave the decision (use exact width/height or just aspect ratio) as far down the stack as possible - so the FE component has full control.

The Iframe Embed paragrah in D8 asks for width/height because that's what iframe embed snippets always contain, but the help text does inform users that only the aspect ratio is guaranteed, not the exact size.


creido-welly (approved):
🚀

PR #383, opened 2021-01-06
#7868 feature: grants awarded search 

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/7868

- add styles for a search box on find grants awarded page

Screenshot 2021-01-06 at 11 16 58

### DIFF
--- a/src/components/SearchForm/_search-form.scss
+++ b/src/components/SearchForm/_search-form.scss
@@ -49,7 +49,10 @@
 }
 
 .search-form__btn {
-  padding: 18px 25px 18px 12px;
+  padding-bottom: calc(2 * var(--space-unit));
+  padding-left: calc(1.5 * var(--space-unit));
+  padding-right: calc(3.125 * var(--space-unit));
+  padding-top: calc(2 * var(--space-unit));
   position: absolute;
   right: 0;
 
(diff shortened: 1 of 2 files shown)


src/components/SearchForm/_search-form.scss
```diff
@@ -43,3 +47,20 @@
...
+  position: absolute;
+  right: 0;
+
+  .btn__icon--right {
+    margin: 0;
+  }
+
+  .btn__text {
+    @extend %visually-hidden;
+  }
```
chriddell: To avoid nesting styles, I think we could add some props to the Button component to allow us to set classnames on the btn__text and btn__icon elements. I noticed we've nested these selectors (btn__icon & btn__text) elsewhere too (e.g. _accordion.scss), so I think it would help fix these issues too.

What do you think? (& @creido-welly, too!)

creido-welly: Reducing nesting is definitely cleaner and keeps separation between components. The Button component already accepts className as a prop so lets use it.

chriddell: I think it may need `textClassName` and `iconClassName` too though, in this instance?

skibinska: @chriddell this is updated now :)

creido-welly (approved):

PR #385, opened 2021-01-07
#7823 Make Contact component handle multiple institutions

aliceh75:
It is now possible to select multiple institutions for a contact in the Drupal backend.

This PR updates the Contact component to accept an array of institutions, rather than a single one.

Note I didn't feel it worthwhile to create an &lt;Institution&gt; component, as this is the only place this is used (and likely if it was used elsewhere, it would be displayed differently).

Matching changes to corporate-react are done and will come once this is merged & published.

### DIFF
--- a/src/components/Contact/Contact.tsx
+++ b/src/components/Contact/Contact.tsx
@@ -11,6 +11,11 @@ type ContactImageSourceProps = {
   image_full_mobile_hi?: string;
 };
 
+type InstitutionProps = {
+  name?: string;
+  country?: string;
+};
+
 type ContactProps = {
   className?: string;
   contactFormLink?: {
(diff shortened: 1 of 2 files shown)


src/components/Contact/Contact.tsx
```diff
@@ -67,13 +70,12 @@ export const Contact = ({
...
       )}
-      {institution && (
-        <p className="cc-contact__institution" itemProp="worksFor">
-          {institutionCountry
-            ? `${institution}, ${institutionCountry}`
-            : institution}
-        </p>
-      )}
+      {institutions &&
+        institutions.map(e => (
```
creido-welly: Minor: you could use object destructuring here to access the keys directly i.e.

```js
institutions.map(({ country, name }) => (
  <p className="cc-contact__institution" itemProp="worksFor">
    {country ? `${name}, ${country}` : name}
  </p>
))}
```

creido-welly (approved):
One minor comment but all good otherwise

src/components/Contact/Contact.tsx
aliceh75: Done ! Note I had to alias `name` to `institutionName` as otherwise it shadowed an outer variable.

PR #386, opened 2021-01-13
#7935 feat: allow TextCard to accept titleAs prop

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/7935

- add `titleAs` prop to the TextCard component

### DIFF
--- a/src/components/TextCard/TextCard.tsx
+++ b/src/components/TextCard/TextCard.tsx
@@ -22,6 +22,7 @@ type TextCardProps = {
   id?: string;
   metaLabel?: string;
   title: string;
+  titleAs?: 'h2' | 'h3';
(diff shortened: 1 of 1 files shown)


src/components/TextCard/TextCard.tsx
```diff
@@ -40,8 +41,10 @@ export const TextCard = ({
   id,
   metaLabel,
   title,
+  titleAs = 'h3',
   type
 }: TextCardProps) => {
+  const Element = titleAs;
```
creido-welly: Call this `TitleElement` to be consistent with other components that use this prop

skibinska: updated :)

creido-welly (approved):

PR #387, opened 2021-01-14
#7923 feat(style): Added sticky positioning to in-page nav

creido-welly:


src/assets/styles/50-utilities/_scaffolding.scss
```diff
@@ -36,5 +36,31 @@
...
+    bottom: -100px;
+    content: '';
+    left: 0;
+    pointer-events: none;
+    position: absolute;
+    right: 0;
+    top: -100px;
+    z-index: var(--z-low);
   }
 }
```
chriddell: I really like this effect and how it's applied... just a thought, would it be clearer if applied via it's own utility selector? `.u-sticky-fade` or something to that effect?

chriddell: Alternatively, a mixin may be appropriate, too. I think this would give it better clarity of purpose & reusability

src/assets/styles/10-settings/_colours.scss
```diff
@@ -58,6 +58,8 @@
   --X-colour-error-red: #d2373f;
   --X-colour-error-pink: #ffebeb;
   --X-colour-dark-blue: #003667;
+
+  --hack-safari-transparent-white: rgba(255, 255, 255, 0.01);
```
chriddell, line 68: What's this for?

src/assets/styles/50-utilities/_scaffolding.scss
chriddell: I feel like we may benefit from these styles being tied to their own specific selector, rather than the `.u-scaffold-grid-item`. I think it makes sense from a readabilty standpoint as what we're doing here feels like a workaround.

What do you think?

chriddell (commented):
I think this is a very good approach given the circumstances (markup & order of components in the page header) and I'm honestly not sure what else we can do in our situation (other than attempting a JS solution).

Would be lovely (although probably frustrating initially!) to take a step back and look at the layout & markup of the entire page header to see if we can absolve ourselves of this workaround, at some point

src/assets/styles/10-settings/_colours.scss
creido-welly: In safari background linear gradients which fade to transparent [display very strangely](https://stackoverflow.com/questions/11829410/css3-gradient-rendering-issues-from-transparent-to-white) - the example specified alpha as 0.001 but 0.01 worked better in my case.

creido-welly: I'll add a code comment to explain that

src/assets/styles/50-utilities/_scaffolding.scss
creido-welly: Makes sense. I've created a utility class and also a mixin. The latter I have included used directly within this file to keep complexity within the [scaffold mapping](https://github.com/wellcometrust/corporate-react/blob/690e8e98aa43da6b7c8eac503d99f490862d8dd1/src/utils/scaffold-mapping.tsx#L153) to minimum.

creido-welly: Mixin used in preference, see comments below

creido-welly:
> Would be lovely (although probably frustrating initially!) to take a step back and look at the layout & markup of the entire page header to see if we can absolve ourselves of this workaround, at some point

Following discussion with Jason, the latest designs lend themselves more to actually placing the nav more or less where it is before it becomes sticky. He did caveat it by saying there are some up and coming design changes which may require the nav to be positioned further up the page initially but that would be the time to review the page header as a whole.

src/assets/styles/10-settings/_colours.scss
chriddell: ace, thanks for that

src/assets/styles/50-utilities/_sticky.scss
```diff
@@ -0,0 +1,11 @@
+// ----------------------------------
```
chriddell, line 1: As we're not using this yet I think we're adding unnecessary bloat to the styles. Should we save it for a future PR? What do you think?

src/assets/styles/50-utilities/_sticky.scss
```diff
@@ -0,0 +1,11 @@
...
+// UI Settings
+// Sticky elements
+// ----------------------------------
+// Design System 1.0 Alpha
+// 2019-11-11
+// ----------------------------------
+
+.u-sticky-fade {
```
chriddell: As we're not using this yet I think we're adding unnecessary bloat to the styles. Should we save it for a future PR? What do you think?

chriddell (approved):
I've left a comment about whether we should be adding the `u-sticky-fade` selector yet as we're not using it, but other than that this looks good to me

PR #389, opened 2021-01-18
#7938 feature: adds status label to Timeline

chriddell:
As per [the latest Funding Scheme page type prototype](https://www.figma.com/file/6x8YZi2UaQ9eRaXnWUy6Iq/Scheme-Page-Design-%26-Prototype), the Timeline component should display a "status label" indicating to users the current state of an application process for a Funding Scheme.

The status label displayed should reflect where the current date sits in the Timeline milestones, therefore it should be calculated on the client to avoid displaying an "outdated" date due to the users viewing a "stale" static render of a page.

See https://github.com/wellcometrust/corporate/issues/7938 for full context.

See also https://github.com/wellcometrust/corporate-react/pull/707 (sibling PR for corporate-react).

## This PR:

- adds `initialStatusLabel` prop to Timeline component
- adds function to Timeline component to calculate most recent milestone's status label (if present)
- updates Timeline.test.tsx to work with updated Component
- updates Timeline.stories.tsx to work with updated Component

## To test:

- check out `https://github.com/wellcometrust/corporate-react/tree/feature/timeline-status-label-%237938`
- `npm link` corporate-components in your local corporate-react
- add `NEXT_PUBLIC_API_DOMAIN=https://wt-corporated8-develop.codeenigma.net` to corporate-react .env
- `npm run dev` in corporate-react
- navigate to https://localhost:3001/what-we-do/our-work/timeline-status-label

## Notes:

The API (corporate-d8) has been updated to return:

- `timeline_status` (string) on the `paragraph--timeline_milestone` include
- `timeline_status` (string) on the `paragraph--timeline` include

You can see the API for an example node [here](https://wt-corporated8-develop.codeenigma.net/api/node/standard/8b370510-0fe2-4dbf-b12e-42a9fe2e9a22?include=sections_ref.content_slices_ref.accordion_slice_ref,sections_ref.content_slices_ref.content_slices_ref.accordion_slice_ref,banner_image_ref.field_image,sections_ref.content_slices_ref.content_slices_ref.institution.parent,sections_ref.content_slices_ref.library_item_ref.paragraphs.content_slices_ref,sections_ref.buttons_ref,sections_ref.content_slices_ref.image_ref.field_image,sections_ref.content_slices_ref.gallery_image_ref.field_image,sections_ref.content_slices_ref.content_slices_ref.linked_file.field_media_file,sections_ref.content_slices_ref.content_slices_ref.linked_file.media_type,sections_ref.content_slices_ref.content_slices_ref.linked_content.listing_image_ref.field_image,sections_ref.content_slices_ref.exclude_from_listing,listing_image_ref.field_image,sections_ref.content_slices_ref.content_slices_ref.content_slices_ref,sections_ref.content_slices_ref.content_slices_ref.image_ref.field_image,sections_ref.content_slices_ref.content_slices_ref.content_slices_ref.linked_content,sections_ref.content_slices_ref.content_slices_ref.content_slices_ref.linked_file.field_media_file,sections_ref.content_slices_ref,sections_ref.library_item_ref.paragraphs.content_slices_ref.content_slices_ref.content_slices_ref,sibling_navigation_ref,sections_ref.content_slices_ref.associated_file,sections_ref.content_slices_ref.timeline_slices_ref,sections_ref.content_slices_ref.video_embed_ref).

### DIFF
--- a/src/components/Timeline/Timeline.stories.tsx
+++ b/src/components/Timeline/Timeline.stories.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import { text } from '@storybook/addon-knobs';
+import { date, text } from '@storybook/addon-knobs';
+import { formatISO } from 'date-fns';
 
 import Timeline from './Timeline';
 import Readme from './Timeline.md';
@@ -12,11 +13,17 @@ const TimelineExample = () => {
   const milestone3ID = 'Milestone 3';
 
   const title = text('title', 'Beatles Albums', generalID);
+  const statusLabel = text(
+    'status label',
+    'Open to new submissions',
+    generalID
+  );
 
   const milestones = [
     {
       body: text('Milestone 1 body', 'This is the body text', milestone1ID),
-      date: text('Milestone 1 date', '1964', milestone1ID),
+      date: formatISO(date('Milestone 1 date', new Date(), milestone1ID)),
+      dateLabel: text('Milestone 1 date label', '1964', milestone1ID),
       linkHref: text(
         'Milestone 1 link href',
         '/beatles-for-sale',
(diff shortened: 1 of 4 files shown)


creido-welly (approved):
As discussed, looks good thus far

creido-welly (approved):
The code looks fine. 

I have one query about users from a different locale which is not GMT e.g. US east coast, would times be read correctly for their particular time zone?

chriddell:
Thanks for asking that @creido-welly - it's made me realise that the date which comes from Drupal is ISO and doesn't contain any locale information (e.g `2021-03-16T23:59:00+00:00`). 

I think we can always assume that this date is in GMT (I'll speak to BE devs and DCPs to confirm), but I think that means we'll need to make it explicit in the code that we're checking the user's locale datetime against a datetime that is not their locale (because it's GMT).

creido-welly:
It's possibly as much a design question as a coding one i.e. how we want time to be represented.

One way to mitigate any misunderstandings could be to include the time zone (GMT, EST etc) in the date label.

lukewhitmore:
Hi Both .. regarding timezone:

image

Drupal provides times according to a system set timezone. 

It's possible to alter this on a per user basis, but we're not using this facility. In any case, as we consume data via one Drupal user account, the dates referenced will always be based from a particular timezone.

chriddell:
Thanks @lukewhitmore - that's interesting. I can see that Drupal provides a UTC date with the timezone adjustment included when I change the region on my local D8.

I changed my region to Vancouver, which meant that this date/time...

Screenshot 2021-01-25 at 14 32 10

...is emitted as...

`2021-01-19T09:00:00-08:00`

...from the API.

This is useful functionality to know I think but as you say we will only ever emit dates from the API in one locale (London), so we just need to make sure the Date that we create on the FE (from the user's locale) is converted to UTC prior to comparison with API (milestone) dates.

chriddell:
After going round the houses for a while on this I've realised that actually we don't need to do anything on the FE.

- The date string from Drupal will come through as UTC (with timezone info)
- new Date(dateString) will run on the client and will return a JS Date with adjusted value for timezone (ie the GMT date string from the API converted to a Date in the user's locale)
- We then use date-fns' [isPast()](https://date-fns.org/docs/isPast) func to find the most recent date, and this also uses the system (user's) locale

Here's some tests I ran by changing my system locale, with a datetime set to 15:00 25th January in the CMS*:

Screenshot 2021-01-25 at 21 12 39 (2)

Screenshot 2021-01-25 at 10 15 47 (2)

Screenshot 2021-01-25 at 15 16 27 (2)

Interestingly my local D8 seems to be set to BST...

PR #400, opened 2021-01-28
Feature/inpage nav interactivity #6682

creido-welly:
- Added configurable 'back to top' link
- Back to top is hidden at mobile breakpoints where the nav does not have position sticky


chriddell (approved):

chriddell:
Nice, I think it makes sense to do this here. I have a feeling going forwards we may make the "on this page" sticky on mobile, so it may even require further changes... I _think_ the original UI for it was sticky on mobile?

(Just checked quickly on Zeplin and have a feeling I'm wrong... think it was the filters on the listing page I was thinking of)

creido-welly:
> Nice, I think it makes sense to do this here. I have a feeling going forwards we may make the "on this page" sticky on mobile, so it may even require further changes... I _think_ the original UI for it was sticky on mobile?

I agree, now it's in place it feels much better.

The code is pretty well placed to accommodate the sticky positioning for mobile too.

PR #401, opened 2021-01-28
#8018 feat: focus state for checkboxes

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8018

- Add focus state for the checkboxes.
- Change hover border colour

I have used `:focus-visible` here and followed [the article about adding backwards compatibility for :focus-visible](https://css-tricks.com/focus-visible-and-backwards-compatibility/)

I decided to use it as I thought it would be a good idea to show focus only for the keyboard navigation, and not on hover.

I tested it in Chrome and Egde and it works as expected.

Please let me know your thoughts! :)

### DIFF
--- a/src/assets/styles/10-settings/_colours.scss
+++ b/src/assets/styles/10-settings/_colours.scss
@@ -24,8 +24,6 @@
   --colour-blue-10: #dbeaff;
   --colour-blue-05: #f0f6ff;
 
-  --colour-concrete-90: #292929;
-
   --colour-cyan-20: #beebf4;
   --colour-cyan-05: #f1fcfd;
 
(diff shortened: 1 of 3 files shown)


src/assets/styles/10-settings/_colours.scss
```diff
@@ -24,8 +24,6 @@
   --colour-blue-10: #dbeaff;
   --colour-blue-05: #f0f6ff;
 
-  --colour-concrete-90: #292929;
```
skibinska, line 27: There were two variables for the same colour (#292929):
`--colour-concrete-90` and `--colour-grey-80`

creido-welly: 👍  it's an older colour name so tnx for removing. Design system colours can be found here FYI
https://app.zeplin.io/project/5b4e06b48ae4580d4871178a/screen/5ea86a03c8963abb60b549cd

creido-welly (approved):

PR #404, opened 2021-02-02
Bug/empty array non rendering #8053

creido-welly:
This PR contains a fix for logic used to conditionally render card or promo authors. Short circuit logic returns the last item which is true so in the case of `authors?.length && <FullWidthPromo ... />` if there are no authors then a 0 is returned as `authors` is an empty array.

It's a minor point but `authors?.length > 0 && (` has been used instead of `!!authors?.length && (`. I believe this is instantly more readable than a forced boolean since the logic is explicitly saying we only render when the number of authors is greater than 0.

### DIFF
--- a/src/components/FullWidthPromo/_full-width-promo.scss
+++ b/src/components/FullWidthPromo/_full-width-promo.scss
@@ -84,14 +84,13 @@
   font-size: var(--body-xs);
   margin: 0;
   text-transform: uppercase;
-}
 
-.cc-full-width-promo__type + .cc-full-width-promo__authors:before {
-  color: var(--colour-amber-30);
-  content: '/';
-  font-size: var(--body-xs);
-  padding-left: var(--space-unit);
-  padding-right: var(--space-unit);
+  &:after {
+    color: var(--colour-amber-30);
+    content: '/';
+    padding-left: var(--space-unit);
+    padding-right: var(--space-unit);
+  }
 }
 
 .cc-full-width-promo__authors {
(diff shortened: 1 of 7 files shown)


src/components/FullWidthPromo/_full-width-promo.scss
```diff
@@ -84,14 +84,13 @@
...
-  font-size: var(--body-xs);
-  padding-left: var(--space-unit);
-  padding-right: var(--space-unit);
+  &:after {
+    color: var(--colour-amber-30);
+    content: '/';
+    padding-left: var(--space-unit);
+    padding-right: var(--space-unit);
+  }
 }
```
chriddell, line 94: This obviously causes the "slash" to appear permanently, even when no authors (image below explains better than words) - but I think you intended this, and I also think it looks better.

Screenshot 2021-02-03 at 15 47 20

chriddell (approved):
Thanks for sorting this, my bad in the first place... I didn't test it properly.

Also agree with your point about not forcing a boolean - I think the method you've used is more appropriate & readable.

creido-welly:
> Thanks for sorting this, my bad in the first place... I didn't test it properly.

NP, my bad too. In Storybook it appeared to be handled correctly but the prop format was subtly different:

Storybook `['']` vs `[]` Drupal API

src/components/FullWidthPromo/_full-width-promo.scss
creido-welly: Yes that was the intention - it makes it consistent with the Image Card and Featured Promo.

PR #405, opened 2021-02-02
News Landing QA bugs

creido-welly:
* Reduced sticky fade overlap
* Enforced specific aspect ratio for ImageCard images to ensure heights are identical (screenshot below)

Screenshot 2021-02-02 at 17 06 51

### DIFF
--- a/src/components/ImageCard/_image-card.scss
+++ b/src/components/ImageCard/_image-card.scss
@@ -40,8 +40,18 @@
 
 .cc-image-card__image {
   @extend %img-responsive;
-
+  @extend %img-cover;
   margin: 0;
+
+  // ensure all card images have the same aspect ratio to enforce the same height
+  @include mq(md) {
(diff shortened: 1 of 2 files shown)


src/components/ImageCard/_image-card.scss
```diff
@@ -40,8 +40,18 @@
 
 .cc-image-card__image {
   @extend %img-responsive;
-
+  @extend %img-cover;
```
creido-welly, line 43: I've stopped short of refactoring this and creating a specific extend as it may well be rendered unnecessary by the following ticket which would ensure placeholders are always the same size as images they may be butting up against in lists
https://github.com/wellcometrust/corporate/issues/8063

chriddell: Thanks for the context Chris, very helpful

chriddell (approved):

PR #408, opened 2021-02-05
 #8083 feature: add sidebar search

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8083

We are using the sidebar search on the grants awarded page and schemes page.

**Where is it used**

The search is a part of filters on the **schemes page**

Screenshot 2021-02-05 at 13 51 22

The search on **[grants awarded page](https://develop.wellcome-corporate-fe.org.uk/grant-funding/people-and-projects/grants-awarded)**

### DIFF
--- a/src/components/SidebarFilter/SidebarFilter.tsx
+++ b/src/components/SidebarFilter/SidebarFilter.tsx
@@ -11,12 +11,11 @@ type SidebarFilterProps = {
     value: string;
   }[];
   allowForceReset?: boolean;
-  children?: JSX.Element | JSX.Element[];
+  children?: JSX.Element;
   className?: string;
   onChange: (value: string) => void;
   onClear: () => void;
   onTagRemove: (value: string) => void;
-  searchComponent?: React.ReactNode;
   tags: {
     name: string;
     items: {
@@ -31,13 +30,12 @@ type SidebarFilterProps = {
 export const SidebarFilter = ({
   activeTags = [],
   allowForceReset,
+  children,
   className,
   onChange,
   onClear,
   onTagRemove,
-  searchComponent,
-  tags,
-  children
+  tags
 }: SidebarFilterProps) => {
   const classNames = cx('cc-sidebar-filter', {
     [className]: className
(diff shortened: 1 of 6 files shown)


src/components/SidebarFilter/SidebarFilter.tsx
```diff
@@ -68,6 +71,7 @@ export const SidebarFilter = ({
           ))}
         </div>
       )}
+      {children}
```
skibinska, line 74: I think this is a clearer solution than adding a new component (SidebarSearch) and it all props.

src/components/SidebarSearch/SidebarSearch.tsx
```diff
@@ -0,0 +1,61 @@
...
+  value
+}: SidebarSearchProps) => {
+  const classNames = cx('cc-sidebar-search', {
+    [className]: className
+  });
+
+  return (
+    <div className={classNames}>
+      <label htmlFor={`search-${id}`} className="cc-sidebar-search__label">
+        {label || 'Search by keyword'}
```
chriddell: Rather than using a conditional, I would suggest setting "Search by keyword" as the default value for the `label` prop.

```
SidebarSearch = ({
  ...otherProps,
  label = 'Search by keyword'
}) => {
```

That way it becomes clearer, earlier that there is a default value for label if you do not pass one.

src/components/SidebarSearch/SidebarSearch.tsx
```diff
@@ -0,0 +1,61 @@
...
+      </label>
+      <input
+        className="cc-sidebar-search__input"
+        name={`search-${id}`}
+        onChange={onChange}
+        onKeyDown={onKeyDown}
+        onKeyUp={onKeyUp}
+        placeholder="e.g. fellowship"
```
chriddell: Would it be useful to have the placeholder as a prop? It may not always be "e.g. fellowship".

src/components/SidebarFilter/SidebarFilter.tsx
chriddell: Agreed

chriddell (changes requested):
Left a couple of minor comments.

Unrelated to this specific PR, I think we should move both of these components to corporate-react. They are very tightly coupled to the wellcome.org website, and I can't imagine they will ever be re-used as they currently are in any other products.

creido-welly (approved):

chriddell (approved):

PR #412, opened 2021-02-09
Image card with CTA component - #8069

DipakShrestha18:


src/components/ImageCardWithCTA/ImageCardWithCTA.test.tsx
```diff
@@ -0,0 +1,19 @@
+// import React from 'react';
```
creido-welly: Is this commented out for a reason?

DipakShrestha18: No, forgot to push it.

DipakShrestha18: @creido-welly ready for re-review.

src/components/ImageCardWithCTA/ImageCardWithCTA.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
```
creido-welly, line 1: There is a lot of duplication here, understandably so, but from a React point of view we should try to make use of _composition_ - or in other words make use of props children to create a base component which has all of the common stuff and into which the markup which is different can be nested.

e.g.

```jsx
<CardBase>
  {description && <RichText>{description}</RichText>}
</CardBase>
```

This is an oversimplification but illustrates the basics. More than happy to chat through this if you like.

Longer term we want to overhaul how listings and cards are composed as it's pretty complex and therefore not ideal. Updating both this component and ImageCard to use the same CardBase will be good step in the right direction and certainly reduce duplication.

Some more info on composition:
https://reactjs.org/docs/composition-vs-inheritance.html
https://epicreact.dev/one-react-mistake-thats-slowing-you-down/

creido-welly (changes requested):
There is a lot of duplication here, understandably so, but from a React point of view we should try to make use of _composition_ - or in other words make use of props children to create a base component which has all of the common stuff and into which the markup which is different can be nested.

e.g.

```jsx
<CardBase>
  {description && <RichText>{description}</RichText>}
</CardBase>
```

This is an oversimplification but illustrates the basics. More than happy to chat through this if you like.

Longer term we want to overhaul how listings and cards are composed as it's pretty complex and therefore not ideal. Updating both this component and ImageCard to use the same CardBase will be good step in the right direction and certainly reduce duplication.

Some more info on composition:
https://reactjs.org/docs/composition-vs-inheritance.html
https://epicreact.dev/one-react-mistake-thats-slowing-you-down/

chriddell:
@creido-welly How about if ImageCard became the "base", and then we had ImageCardWithMetaData and ImageCardWithCTALink as the "extensions"... 🤯

creido-welly:
> @creido-welly How about if ImageCard became the "base", and then we had ImageCardWithMetaData and ImageCardWithCTALink as the "extensions"...

Sure if it makes sense to do that. ImageCard as it stands is effectively ImageCardWithMetaData.

DipakShrestha18:
> > @creido-welly How about if ImageCard became the "base", and then we had ImageCardWithMetaData and ImageCardWithCTALink as the "extensions"...
> 
> Sure if it makes sense to do that. ImageCard as it stands is effectively ImageCardWithMetaData.

Do you mean extending the existing component?

chriddell:
Easiest way to think about it is probably:

- Renaming ImageCard -> ImageCardWithMetaData
- Copying all of the code from ImageCardWithMetaData to a new component: ImageCard
- Refactoring ImageCard to:
  - remove the parts which are unique to ImageCardWithMetaData and ImageCardWithCTALink 
  - add a `children` prop
- Refactoring ImageCardWithMetaData and ImageCardWithCTALink to use ImageCard and passing unique parts as `children`


Basically we would have something similar to*:

```js
const ImageCard = ({ children, ...otherProps }) => (
  <article className="c-image-card">
    <figure className="c-image-card__img">
      <img src={src} />
    </figure>
    {children}
  </article>
);

const ImageCardWithMetaData = ({ props }) => (
  <ImageCard>
    <div className="c-image-card__meta">
      <p>{props.date}</p>
      <p>{props.author}</p>
    </div>
  </ImageCard>
);

const ImageCardWithCTALink = ({ props }) => (
  <ImageCard>
    <a className="c-image-card__cta" href={props.href}>
  </ImageCard>
);
```

*this is VERY simplified

DipakShrestha18:
Should it make sense to work that as separate job? It feels like we're side tracking as Yssy wants to get this component out this sprint. What's your view?

I can create a ticket for this and work on it next week.

creido-welly:
@DipakShrestha18 since time is short I have no objection as long as it does get done

creido-welly (approved):

DipakShrestha18:
I have create github ticket for refactoring and consistent naming. 
https://github.com/wellcometrust/corporate/issues/8108

PR #414, opened 2021-02-10
Feat/description list component #8092

lukewhitmore:
Hi .. issue for this is [here](https://github.com/wellcometrust/corporate/issues/8092).

I've created another branch in `corporate_react` containing the implementation for this (same name as this branch - PR [here](https://github.com/wellcometrust/corporate-react/pull/761)).

### DIFF
--- a/src/components/DescriptionList/DescriptionList.tsx
+++ b/src/components/DescriptionList/DescriptionList.tsx
@@ -11,14 +11,13 @@ export type DescriptionProps = {
 
 type DescriptionListProps = {
   items: DescriptionProps[];
-  id: string;
 };
 
 export const DescriptionList = ({ items }: DescriptionListProps) => (
   <div className="cc-description-list">
     <dl>
       {items
-        .filter(({ markup }) => markup !== null)
+        .filter(({ markup }) => markup !== '')
         .map(({ title, markup }) => (
           <Fragment key={title}>
             <dt className="cc-description-list__title">{title}</dt>
--- a/src/components/DescriptionList/DescriptionList.tsx
+++ b/src/components/DescriptionList/DescriptionList.tsx
@@ -14,7 +14,7 @@ type DescriptionListProps = {
   id: string;
 };
 
-export const DescriptionList = ({ items, id }: DescriptionListProps) => (
+export const DescriptionList = ({ items }: DescriptionListProps) => (
   <div className="cc-description-list">
     <dl>
       {items
--- a/src/components/DescriptionList/DescriptionList.tsx
+++ b/src/components/DescriptionList/DescriptionList.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 
 import { parseHtml } from 'utils/parse-html';
 
@@ -16,16 +16,16 @@ type DescriptionListProps = {
 
 export const DescriptionList = ({ items, id }: DescriptionListProps) => (
   <div className="cc-description-list">
-    <dl key={id}>
+    <dl>
       {items
-        .filter(({ markup }) => markup !== null )
+        .filter(({ markup }) => markup !== null)
         .map(({ title, markup }) => (
-          <>
+          <Fragment key={title}>
             <dt className="cc-description-list__title">{title}</dt>
             <dd className="cc-description-list__definition">
               {parseHtml(markup)}
             </dd>
-          </>
+          </Fragment>
         ))}
     </dl>
   </div>
--- a/src/components/DescriptionList/DescriptionList.tsx
+++ b/src/components/DescriptionList/DescriptionList.tsx
@@ -1,0 +1,34 @@
+import React from 'react';
+
+import { parseHtml } from 'utils/parse-html';
+
+export type DescriptionProps = {
+  title: string;
+  markup: string;
+  url: string;
+  name: string;
+};
+
+type DescriptionListProps = {
+  items: DescriptionProps[];
+  id: string;
+};
+
+export const DescriptionList = ({ items, id }: DescriptionListProps) => (
+  <div className="cc-description-list">
+    <dl key={id}>
+      {items
+        .filter(({ markup }) => markup !== null )
+        .map(({ title, markup }) => (
+          <>
+            <dt className="cc-description-list__title">{title}</dt>
+            <dd className="cc-description-list__definition">
+              {parseHtml(markup)}
+            </dd>
+          </>
+        ))}
+    </dl>
+  </div>
+);
+
+export default DescriptionList;
--- a/src/components/DescriptionList/_description-list.scss
+++ b/src/components/DescriptionList/_description-list.scss
@@ -10,25 +10,4 @@
 .cc-description-list__definition {
   display: block;
   margin: 0 0 var(--space-sm) 0;
-
-  a {
-    @extend %anchor-base;
-  }
-
-  p {
-    margin: 0;
-  }
-
-  span {
-    display: inline;
-  }
-
-  span:after {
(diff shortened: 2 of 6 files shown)


src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,34 @@
...
+  id: string;
+};
+
+export const DescriptionList = ({ items, id }: DescriptionListProps) => (
+  <div className="cc-description-list">
+    <dl key={id}>
+      {items
+        .filter(({ markup }) => markup !== null )
+        .map(({ title, markup }) => (
+          <>
```
creido-welly: You will need to add a key to the fragment since this is the list of mapped items i.e. `<Fragment key={title}>`.

src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,34 @@
...
+};
+
+type DescriptionListProps = {
+  items: DescriptionProps[];
+  id: string;
+};
+
+export const DescriptionList = ({ items, id }: DescriptionListProps) => (
+  <div className="cc-description-list">
+    <dl key={id}>
```
creido-welly: I suspect the key here is not needed

src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,34 @@
...
+  url: string;
+  name: string;
+};
+
+type DescriptionListProps = {
+  items: DescriptionProps[];
+  id: string;
+};
+
+export const DescriptionList = ({ items, id }: DescriptionListProps) => (
```
creido-welly: `<dl>` is a **definition** list. Should this be `DefinitionList` or was DescriptionList intentional?

lukewhitmore: Hi .. this was intentional!

I considered `DefinitionList` but thinking about it, it felt like `DescriptionLIst` was more appropriate as we're providing a specific visual treatment, which happens to use `<dl>` elements semantically.

src/components/DescriptionList/DescriptionList.tsx
lukewhitmore: This has now been changed! Thx

creido-welly (approved):

src/components/DescriptionList/_description-list.scss
```diff
@@ -0,0 +1,34 @@
+.cc-description-list__title {
+  display: block;
+  float: left;
+  font-size: var(--body-lg);
+  font-weight: bold;
+
+  margin: 0 var(--space-sm)  0 0;
```
chriddell: Remove (one of) the spaces after the margin right value (there is a double space)

chriddell: I feel like prettier **should** catch this, but it doesn't, which is annoying

src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,34 @@
...
```
chriddell: `id` is never used, remove if so.

src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,34 @@
...
+  <div className="cc-description-list">
+    <dl>
+      {items
+        .filter(({ markup }) => markup !== null)
```
chriddell: If markup can be (typeof) `null` then it should be marked as optional in the relevant type alias.

```
markup?: string
```

Also might be worth thinking about what to do in the condition where the string is empty

src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,34 @@
...
+        .map(({ title, markup }) => (
+          <Fragment key={title}>
+            <dt className="cc-description-list__title">{title}</dt>
+            <dd className="cc-description-list__definition">
+              {parseHtml(markup)}
```
chriddell, line 25: Personally I would use the `RichText` component here as it both parses the HTML for us and allows to do some global transforms (e.g. adding selectors, preventing script tags being inserted).

Can understand there may be times not to use this, so fair enough if this is one of those!

src/components/DescriptionList/_description-list.scss
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  span:after {
+    content: ', ';
+    display: inline;
+  }
+
+  span:last-child:after {
+    content: '';
+  }
```
chriddell: Is the nesting necessary due to "legacy" markup being migrated from D7?

src/components/DescriptionList/_description-list.scss
```diff
@@ -0,0 +1,34 @@
+.cc-description-list__title {
+  display: block;
+  float: left;
+  font-size: var(--body-lg);
+  font-weight: bold;
+
```
chriddell, line 6: Unnecessary line break here (again, Prettier should catch this 😩, sorry!)

src/components/DescriptionList/DescriptionList.tsx
lukewhitmore: Hey - I did consider that, but it added a nested `div` which felt unnecessary / didn't read well semantically.

Bearing this in mind, I took the `parseHtml` function which as far as I can tell performs santisation. Would definitely be possible to run this through `formatHTMLString` as well if you think it's worth it?

One thing to bear in mind is that the links need to open in the same window.

chriddell: If it's just the `<div>` element that puts you off using it then I think

chriddell: I don't think we need to worry about an extraneous `<div>` element - the `dd` element can contain any flow content as children (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dd), so I don't think we should worry about a div.

Re: links in same window - this is something which needs to change globally, so I would say still use the RichText in this instance because we'll be making a change to RichText soon enough that ensures `<a>` tags with external URLs are no longer automatically marked up with `target="_blank" rel="noopener noreferrer"`.

lukewhitmore: Sure, that makes good sense.

Just so you're aware; there are two types of content that corporate-react is going to place here ..

image

* The first type is a list of terms; and can be seen in the top three items in the screenshot.
* The second type is a block of rich text; and can be seen in the bottom two items.

lukewhitmore: Bearing this in mind, I'm wondering whether both should be displayed within a `RichText` block. And if so, perhaps I should be inserting a comma character for list item via corporate-react, rather than adding the comma using a pseudo element?

chriddell: That is helpful thanks.

So actually only the first type is markup, and the second type is a text string (though both ultimately come through as strings?).

My thought is that we could use RichText regardless, and add a className of (e.g.) `u-dis-i` - a utility selector which sets `display: inline`. I think this would work, though could be wrong.

Another way would be adding a className of (e.g.) `c-description-list__rich-text` and hooking any relevant style rules to that.

src/components/DescriptionList/_description-list.scss
lukewhitmore: I think I'll be able to remove this styling if I used the `RichText` component.

chriddell: 😎

chriddell:
Superseded by https://github.com/wellcometrust/corporate-components/pull/422

PR #416, opened 2021-02-15
#8120: feat add select to ResultsCount

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8120

- add select to ResultsCount component
- used on the `Find a scheme` page

Screenshot 2021-02-15 at 14 54 05

### DIFF
--- a/src/components/ResultsCount/ResultsCount.tsx
+++ b/src/components/ResultsCount/ResultsCount.tsx
@@ -1,17 +1,29 @@
 import React from 'react';
 import cx from 'classnames';
+import Label from '../Label';
 
 type ResultsCountProps = {
   className?: string;
   currentCount: string;
+  id?: string;
+  onChange?: () => void;
+  options?: {
+    label: string;
+    value: string;
+  }[];
   resultsCount: number;
-  sortedBy: string;
+  selectedValue?: string;
+  sortedBy?: string;
 };
 
 export const ResultsCount = ({
   className,
   currentCount,
+  id,
+  onChange,
+  options,
   resultsCount,
+  selectedValue,
   sortedBy
 }: ResultsCountProps) => {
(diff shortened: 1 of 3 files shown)


src/components/ResultsCount/ResultsCount.tsx
```diff
@@ -26,7 +38,30 @@ export const ResultsCount = ({
           {currentCount} results of {resultsCount}
         </strong>
       </p>
-      <p className="cc-results-count__sort">Sorted by {sortedBy}</p>
+      {sortedBy && (
```
creido-welly: There is a possibility of sortedBy text and the `options` select being rendered at the same time. Is it worth having some logic here to decide which one is displayed by preference? Would make more sense to prefer the select I think if sort options are available.

src/components/ResultsCount/ResultsCount.tsx
```diff
@@ -26,7 +38,29 @@ export const ResultsCount = ({
...
+            value={selectedValue}
+          >
+            {options?.map(option => (
+              <option key={`option-${option.label}`} value={option.value}>
+                {option.label}
+              </option>
+            ))}
+          </select>
+        </div>
+      ) : (
```
creido-welly, line 61: It should keep the logic for checking presence of sortedBy 

```js
  // ...
  ) : sortedBy && (
  // ...
```

creido-welly: Minor: I don't think the outer parens (brackets) are needed

creido-welly (approved):
One minor comment re extraneous parens, otherwise all good

src/components/ResultsCount/ResultsCount.tsx
skibinska: I have just checked and it is added by prettier.

PR #420, opened 2021-02-17
#8128 feat: add Checkbox component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8128

- Add checkbox component
- Update SidebarFilter with a new checkbox component

I built `Checkbox` in the same way as [RadioInput](https://github.com/wellcometrust/corporate-components/blob/develop/src/components/RadioInput/RadioInput.tsx)

Screenshot 2021-02-17 at 14 03 10


src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const Checkbox = forwardRef(
+  (
+    {
+      checked,
+      className,
+      describedBy,
+      disabled,
+      id,
+      isInvalid,
```
skibinska, line 29: As I mentioned in the PR description I built this component in the same way as `RadioInput`. I am not sure when we would like to use `isInvalid` prop. Is it the same as `disabled` or is it related to an error message? Plus we don't have styles for `cc-checkbox--is-invalid` nor `cc-radio-input--is-invalid`. Should we keep it then?

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,78 @@
...
+      'cc-checkbox--is-invalid': isInvalid,
+      [className]: className
+    });
+
+    return label ? (
+      <div className={classNames}>
+        <input
+          aria-describedby={describedBy}
+          checked={checked}
+          className="cc-checkbox__input-element"
```
creido-welly: Minor point but it'll save a little file size if you rename this to `cc-checkbox__input` since there are a fair number of style rules related to it

src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,78 @@
...
+          ref={ref}
+          required={isRequired}
+          type="checkbox"
+          value={value}
+        />
+        <Label text={label} htmlFor={id} className="cc-checkbox__label" />
+        <Icon className="cc-checkbox__icon" name="checkmark" />
+      </div>
+    ) : (
+      <input
```
creido-welly: Just out of interest, are you using this anywhere to render an unlabelled checkbox? I can't imagine any scenario where you would want to unless it was labelled more remotely.

src/components/Checkbox/Checkbox.tsx
creido-welly: `isInvalid` would indicate to me that there is an error with this input - it should/should not be checked given certain conditions

creido-welly: I realise I didn't answer the question - I'd say it definitely belongs with the component. Perhaps create another ticket to address invalid styles and prompt the designers for some guidelines or creative.

src/components/Checkbox/Checkbox.tsx
skibinska: I was just following the RadioInout example. Happy to remove it.

skibinska:
@creido-welly this is ready to review :)

creido-welly (approved):

PR #425, opened 2021-02-19
Fix/refactoring stats card fact card

DipakShrestha18:
Refactoring stastisticcard to FactCardWithHyperlink.


chriddell:
Branch should be named `bug/{branch_name}` (rather than `fix/`) - sorry, very annoying point but trying to maintain consistency as much as possible so it's easier to trace if things go wrong!

chriddell (changes requested):
Branch should be named `bug/{branch_name}` (rather than `fix/`) - sorry, very annoying point but trying to maintain consistency as much as possible so it's easier to trace if things go wrong!

You should be able to just check out a new branch with a different name from this one, and create a new PR? So hopefully fairly straightforwards.

DipakShrestha18:
> Branch should be named `bug/{branch_name}` (rather than `fix/`) - sorry, very annoying point but trying to maintain consistency as much as possible so it's easier to trace if things go wrong!
> 
> You should be able to just check out a new branch with a different name from this one, and create a new PR? So hopefully fairly straightforwards.

Will do.

PR #426, opened 2021-02-19
#8082 bug: refactoring stats card fact card

DipakShrestha18:


chriddell (changes requested):
Code looks good, but... can we name the FE component `FactCard` and drop the `WithHyperlink`? Reasons being...

- we do not have two variants `FactCard` and `FactCardWithHyperlink`
- `WithHyperlink` is fairly meaningless (the term "Hyperlink" is deprecated and doesn't tend to be used any more)
- the relationship between `FactCard` and `fact_card_with_hyperlink` is made explicit elsewhere in the code, so people are free to use either term and the relationship is visible
- it is what I proposed here https://github.com/wellcometrust/corporate/issues/8082#issuecomment-774061301 (this is the least important, really)

chriddell (commented):
Sorry, another annoying point... the branch name should have a ticket number on the end. Also, just a thought - I don't think this is a `bug` branch, nothing is broken (it's just that we use the wrong names).

I think it should be a feature, so

`feature/refactoring-statistic-card-to-fact-card-#8082`

src/components/FactCard/FactCard.tsx
```diff
@@ -13,29 +13,26 @@ type StatisticCardProps = {
   titleAs?: 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 };
 
-export const StatisticCard = ({
+export const FactCardProps = ({
```
creido-welly: You have duplication of the type `FactCardProps` here.  I'm guessing this is a find and replace oversight. It should be `FactCard`.

src/components/FactCard/FactCard.tsx
```diff
@@ -54,4 +51,4 @@ export const StatisticCard = ({
   );
 };
 
-export default StatisticCard;
+export default FactCardProps;
```
creido-welly: See comment on line 16 - default export should be `FactCard` to match

src/components/FactCard/FactCard.tsx
DipakShrestha18: You're right find and replace. 

Re-review please.

creido-welly (approved):

DipakShrestha18:
@chriddell Not sure what's your CR is in this. It's not allowing me to merge this PR.

chriddell:
This is mostly fine now, though the branch name should have the ticket number on the end.

Can you check out a new branch (or rename this one - not sure if that's actually possible...):

```feature/refactoring-statistic-card-fact-card-#<ticketNumber>```

(I wouldn't consider this a bug - it's a refactoring but nothing is broken)

PR #434, opened 2021-03-05
Bug/8263 ensure label is inline

lukewhitmore:
Small fixes for `DescriptionList` component.

Issue [here](https://github.com/wellcometrust/corporate/issues/8263)

### DIFF
--- a/src/components/DescriptionList/_description-list.scss
+++ b/src/components/DescriptionList/_description-list.scss
@@ -9,5 +9,9 @@
 
 .cc-description-list__definition {
   display: block;
-  margin: 0 0 var(--space-sm) 0;
+  margin: 0 0 var(--space-md) 0;
+}
+
+.cc-description-list__definition-inner {
+  display: inline;
 }
--- a/src/components/DescriptionList/DescriptionList.tsx
+++ b/src/components/DescriptionList/DescriptionList.tsx
@@ -3,8 +3,7 @@
  * of terms and their associated data.
  */
 import React, { Fragment } from 'react';
-
-import { parseHtml } from 'utils/parse-html';
+import RichText from 'RichText';
 
 type DescriptionListProps = {
   items: {
@@ -36,7 +35,9 @@ export const DescriptionList = ({ items }: DescriptionListProps) => (
           <Fragment key={title}>
             <dt className="cc-description-list__title">{title}</dt>
             <dd className="cc-description-list__definition">
-              {parseHtml(markup)}
+              <RichText className="cc-description-list__definition-inner">
+                {markup}
+              </RichText>
             </dd>
           </Fragment>
         ))}


creido-welly (approved):

src/components/DescriptionList/_description-list.scss
```diff
@@ -9,5 +9,9 @@
 
 .cc-description-list__definition {
   display: block;
-  margin: 0 0 var(--space-sm) 0;
+  margin: 0 0 var(--space-md) 0;
```
chriddell, line 12: The issue doesn't mention the margins, so just checking this change was intentional?

src/components/DescriptionList/_description-list.scss
```diff
@@ -9,5 +9,9 @@
...
+}
+
+.cc-description-list__definition-inner {
+  display: inline;
 }
```
chriddell, line 17: I'm not 100% sure this is necessary. See screenshot (happy to be proven wrong!)

Screenshot 2021-03-10 at 14 43 23

src/components/DescriptionList/_description-list.scss
lukewhitmore: Hi .. yes, this is intentional.

src/components/DescriptionList/_description-list.scss
lukewhitmore: Without this, the display breaks when longer blocks of text run into multiple lines.

chriddell: Good point.

Fwiw, I think using `float` is the wrong approach. It takes the element out of the flow and causes the need for more code to rectify (e.g. `clear: both`).

Would suggest using `display: inline` with a wrapping div. Will avoid layout issues I think.

```html
<dl>
  <div>
    <dt style="display: inline">term title</dt>
    <dd style="display: inline"><div class="cc-rich-text">term data</div></dd>
  </div>
 </dl>
```

Screenshot 2021-03-10 at 15 49 06

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#wrapping_name-value_groups_in_htmlelementdiv_elements

lukewhitmore: Hey @chriddell - finally got round to looking at this.

Re. float, I see your point, but I'd argue the float doesn't need clearing as it's proceed by a block level level element that isn't floated. 

I've just tried to run with what you're suggesting and it's not providing the layout required.

If it's okay with you, I'd rather go with the current approach. While the HTML5 spec does allow a `div` to be added here, it feels that we're using deliberately adding markup to influence the styling.

Would you be okay with this?

chriddell: No worries - if you've tried the approach I suggested and still think yours is better, am totally happy for you to go ahead with that

chriddell (approved):

PR #435, opened 2021-03-08
Feature 8272: On text cards, title may not be present

stuartwilkes:
Relates https://github.com/wellcometrust/corporate/issues/8272.

Makes display of title element conditional on the presence of a title to handle "No results" behaviour without an empty element.

### DIFF
--- a/src/components/TextCard/TextCard.tsx
+++ b/src/components/TextCard/TextCard.tsx
@@ -84,15 +84,17 @@ export const TextCard = ({
           )}
         </div>
       )}
-      <TitleElement className="cc-text-card__title">
-        {type === 'file' ? (
-          parseHtml(title)
-        ) : (
-          <a href={href} className="cc-text-card__link">
-            {parseHtml(title)}
-          </a>
-        )}
-      </TitleElement>
+      {title && (
+        <TitleElement className="cc-text-card__title">
+          {type === 'file' ? (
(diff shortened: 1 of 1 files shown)


creido-welly (approved):
Code is absolutely fine, just a query about when the circumstances of this happening. Presumably you've actually come across issues with a title not being available?

The ticket gives an example of a dynamic listing for which no results are returned but looking at that code, the list and therefore cards shouldn't render anyway.

stuartwilkes:
> Code is absolutely fine, just a query about when the circumstances of this happening. Presumably you've actually come across issues with a title not being available?
> 
> The ticket gives an example of a dynamic listing for which no results are returned but looking at that code, the list and therefore cards shouldn't render anyway.

So it relates to this: https://github.com/wellcometrust/corporate-react/pull/821

Firstly, there's actually a bug - currently, the behaviour isn't that the cards don't render - the listing remains displaying "Fetching data" due to the way the check is performed. 

Secondly, we want to be able to return custom "No results" text in some instances - especially for "Jobs".

In that instance, we want to display a card that displays "No results" - and it doesn't really make sense to have a title.

I guess we could also handle it with a discrete "No results" component? But this seemed a better way to handle it...

creido-welly:
That makes sense. A text card should be fine for now - we can always change it to something else further down the line

PR #436, opened 2021-03-10
8302 feature: adds CTALink component

chriddell:
# Description

To avoid repetition in styles I have modularised ("converted to a component") the "CTA Link" style that was previously used only within this repo. It is now being exported to be used by consuming applications, as well as being used by internal components such as FullWidthPromo and FactCard.

This PR makes the necessary changes to modularise CTALink.

See wellcometrust/corporate/issues/8302 for more information.

# To test (locally)

- `git pull {this_branch}`
- `npm run storybook`
- Open "CTA Link"
- Check the styles work as expected
- Open "FactCard"
- Check the styles work as expected
- Open "FillWidthPromo"
- Check the styles work as expected

### DIFF
--- a/src/components/CTALink/CTALink.tsx
+++ b/src/components/CTALink/CTALink.tsx
@@ -1,0 +1,45 @@
+/**
+ * @file CTA Link component; wrapper for Button component with updated styles
+ * and which should only be used when HTML element should be an <a> tag.
+ */
+
+import React from 'react';
+import cx from 'classnames';
+
+import Button from 'Button';
+
+type CTALinkProps = {
+  className?: string;
+  href: string;
+  text: string;
+};
+
+/**
+ * CTALink component; wraps the Button element and always returns
+ * an `<a>` tag with "CTA Link" styles.
+ *
+ * @param {string} className
+ * @param {string} href
+ * @param {string} text
+ *
+ * @returns {ReactElement}
+ */
+export const CTALink = ({ className, href, text }: CTALinkProps) => {
+  const classNames = cx('cc-cta-link', {
+    [className]: className
+  });
+
+  return (
+    <Button
+      className={classNames}
+      href={href}
+      icon="arrowLong"
+      iconPlacementSwitch
+      variant="unstyled"
+    >
+      {text}
+    </Button>
+  );
+};
+
+export default CTALink;
--- a/src/components/CTALink/CTALink.tsx
+++ b/src/components/CTALink/CTALink.tsx
@@ -2,7 +2,6 @@
  * @file CTA Link component; wrapper for Button component with updated styles
  * and which should only be used when HTML element should be an <a> tag.
  */
-
 import React from 'react';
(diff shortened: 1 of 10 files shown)


src/components/CTALink/CTALink.tsx
```diff
@@ -0,0 +1,45 @@
...
+import Button from 'Button';
+
+type CTALinkProps = {
+  className?: string;
+  href: string;
+  text: string;
+};
+
+/**
+ * CTALink component; wraps the Button element and always returns
```
creido-welly, line 17: This JSdoc could be combined with the `@file`. I don't think you need to document the props since **type CTALinkProps** is effectively doing that - we also don't do this for any other component

chriddell: I think it's good practice to include both a top-level `@file` jsdoc as well as a comment declaring the individual component's purpose. I think this will stand us in good stead should we want to add more components/functions within the same module file, and also maintains consistency in files which have multiple exports and those which have single exports.

There are other components/files which do this (see Timeline.tsx, FormFieldError.tsx, Link.tsx, header-nav.ts, Video.tsx among others).

I find documenting the props (in addition to their type aliases) can be helpful if there is extra context that we can provide, e.g. in the instance that we expect a HTML string, or where the purpose of the prop may not be clear enough.

I agree that our jsdoc commenting is somewhat haphazard though and we would benefit from some explicit conventions (enforced either via the linter or by each other in our PRs!).

creido-welly (approved):
🚀 nice pattern spotting

src/components/CTALink/CTALink.tsx
creido-welly: I stand corrected, although taking the example of FieldErrors the JSdoc params aren't adding anything which isn't already given by the type definition.

```jsx
type FormFieldErrorProps = {
  className?: string;
  errors: string | {};
  id: string;
};

/**
 * Renders a list or paragraph of error messages
 *
 * @param {string} className
 * @param {string|object} errors
 * @param {string} id
 */
export const FormFieldError = ({
  className,
  errors,
  id
}: FormFieldErrorProps) => {
```

Documentation is important and it's definitely good practice to have conventions for it, but for simpler components, particularly when we have one file one component @file _and_ component docs _with_ @params _and_ type as well isn't necessarily needed and can even just be duplication. More complex components and/or multiple components in one file certainly. For those reasons I think guidelines and then common sense in reviewing PRs would serve us well. In terms of linting, warnings are fine but errors are prob a step too far.

chriddell: Yeah I can't argue with that, it's definitely duplication... but is it worthy duplication? By duplicating param docs (type alias & jsdoc comment) we gain consistency in how our components are written, so....

Is the question: duplication vs. consistency?

creido-welly: Sure it's worthy. We can take this debate offline - get it merged!

PR #437, opened 2021-03-11
8293 task: scheme finder design updates

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8293

Implement design QA

### This PR:

1. add story for ResultsCount
2. two versions of  ResultsCount have now the same spacing

3. remove margin-bottom from the page title component https://github.com/wellcometrust/corporate/issues/8279#issuecomment-795367140
4. update styles for filters

Screenshot 2021-03-11 at 09 08 35

### Test

1. fetch and checkout this branch: `task/8293-scheme-finder-design-updates`
2. go to http://localhost:3001/grant-funding/schemes

### Updated scheme finder page:

Screenshot 2021-03-11 at 09 13 14

### DIFF
--- a/src/components/ResultsCount/ResultsCount.tsx
+++ b/src/components/ResultsCount/ResultsCount.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { ChangeEvent } from 'react';
 import cx from 'classnames';
 import Label from '../Label';
 
@@ -6,7 +6,7 @@ type ResultsCountProps = {
   className?: string;
   currentCount: string;
   id?: string;
-  onChange?: (event: React.ChangeEvent<HTMLSelectElement>) => void;
+  onChange?: (event: ChangeEvent<HTMLSelectElement>) => void;
   options?: {
     label: string;
     value: string;
--- a/src/components/ResultsCount/ResultsCount.tsx
+++ b/src/components/ResultsCount/ResultsCount.tsx
@@ -6,7 +6,7 @@ type ResultsCountProps = {
   className?: string;
   currentCount: string;
   id?: string;
-  onChange?: () => void;
+  onChange?: (event: React.ChangeEvent<HTMLSelectElement>) => void;
   options?: {
     label: string;
     value: string;
--- a/src/components/SidebarFilter/_sidebar-filter.scss
+++ b/src/components/SidebarFilter/_sidebar-filter.scss
@@ -76,3 +76,19 @@
     color: var(--colour-blue-60);
   }
 }
+
+.cc-sidebar-filter__checkbox {
+  font-size: var(--body-md);
+
+  .cc-checkbox__label {
+    padding-bottom: var(--space-md);
+  }
+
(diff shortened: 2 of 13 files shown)


src/components/SidebarFilter/_sidebar-filter.scss
```diff
@@ -1,5 +1,15 @@
 // .cc-sidebar-filter {}
 
+.cc-sidebar-filter__body {
+  .cc-accordion__button {
+    font-size: var(--body-md);
+  }
+
+  .cc-accordion__button .icon {
+    color: var(--colour-black);
+  }
```
skibinska: An accordion as a filter has different styles: smaller font-size, black plus icon, updated styles for checkboxes (see: https://github.com/wellcometrust/corporate-components/pull/437/files#diff-f7e6b45f11fc8c5b6939f03c7dd6746e385c590188bb9cd9808f2abdb4b276f6R78).

Maybe a better idea is to add a prop to the Accordion component?

src/components/ResultsCount/ResultsCount.tsx
```diff
@@ -6,7 +6,7 @@ type ResultsCountProps = {
   className?: string;
   currentCount: string;
   id?: string;
-  onChange?: () => void;
+  onChange?: (event: React.ChangeEvent<HTMLSelectElement>) => void;
```
creido-welly: Instead of React.ChangeEvent use destructuring in the import `import React, { ChangeEvent } from 'react';` and change this to `event:ChangeEvent<HTMLSelectElement>`

src/components/SidebarFilter/_sidebar-filter.scss
creido-welly: Yes I agree. A string union would be a good type to use to allow for other variants in future i.e.

```
// type definition
variant: 'standard' | 'filter';
// ...

// prop
variant = 'standard',
// ...
```

src/components/SidebarFilter/_sidebar-filter.scss
```diff
@@ -33,6 +43,7 @@
 .cc-sidebar-filter__tags-list-item {
   background-color: var(--colour-blue-05);
   color: var(--colour-blue-60);
+  font-size: calc(12/16 * var(--text-base-size)); // we don't have this size in _typography.scss
```
creido-welly: You could always specify this as a CSS variable either scoped globally in `:root` (in this file still) or locally in this block

Locally scoped variables is something I'm still getting my head around but here's some further info:
https://www.freecodecamp.org/news/everything-you-need-to-know-about-css-variables-c74d922ea855/#scoping-css-variables

creido-welly (approved):
Codewise I think it's fine. I'm unable to test locally as I'm getting npm link issues. Two minor comments but nothing to stop this from being merged that I can see.

skibinska:
@creido-welly I have implemented your suggestions. Would it be okay to have a look at these updates when you have a moment? :)

creido-welly (approved):

PR #441, opened 2021-03-12
8315 task(style): generates suite of grid-cell selectors

chriddell:
# Context

As we update the UI design of our website it is necessary to move away from the strict 3 column grid we are currently using.

# Description

This PR adds more fine-tuned control of grid cells by adding a suite of selectors that will allow a developer to set grid cells to span whatever columns they desire, e.g.

`.grid__cell--1-to-6`
`.grid__cell--1-to-13`
`.grid__cell--6-to-13`

etc.

Once this is released it will be necessary to update the usage of `<GridCell />` in consuming applications to ensure the behaviour of `<GridCell colStart={4} colEnd={12}>` stays the same (12 will need to become 13).

# Test

- `git pull {this_branch}`
- `npm run build`
- `open dist/style.css`
- Open the `dist/style.css` file in your IDE
- Find `grid__cell--1-to-2`
- Inspect selectors/style are as expected

### DIFF
--- a/src/components/Grid/Cell/_grid-cell.scss
+++ b/src/components/Grid/Cell/_grid-cell.scss
@@ -76,13 +76,6 @@
   }
 }
 
-// custom column widths
-.grid__cell--4-to-12 {
-  @include mq(md) {
-    @include grid-column(4, 13);
-  }
-}
-
 /**
  * Creates a suite of grid selectors to provide granular
  * control over the exact columns a grid cell should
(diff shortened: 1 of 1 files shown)


src/components/Grid/Cell/_grid-cell.scss
```diff
@@ -76,9 +76,31 @@
   }
 }
 
-// custom column widths
-.grid__cell--4-to-12 {
-  @include mq(md) {
-    @include grid-column(4, 13);
+/**
```
creido-welly, line 79: HUZZAH 📯

creido-welly (approved):
🎠

PR #442, opened 2021-03-15
8314 task: move files to CR

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8314

### This PR

- removes  `.scaffolding.scss`
- removes `sticky-fade` mixin (I checked it was used ones in `.scaffolding.scss`)
- adds `_hacks.sccs` file
- there are now in CR


src/assets/styles/50-utilities/_scaffolding.scss
```diff
@@ -40,3 +40,9 @@
     @include grid-column(1, 13);
   }
 }
+
+.u-scaffold-grid-item--extended {
+  @include mq(md) {
+    @include grid-column(4, 13);
+  }
+}
```
chriddell: I'm very surprised to find these styles in corporate-components. I feel they should be declared in corporate-react as the component they apply to only exists in corporate-react.

What do you think @creido-welly?

creido-welly: Yes it makes sense to move them

chriddell (approved):
The best kind of PR = one where there only deletions 💫

chriddell (approved):
👍

PR #443, opened 2021-03-16
8287 task: add search pane

creido-welly:
* Added new close icon for site header

Design - https://zpl.io/2jgyoWO


src/components/Button/Button.stories.tsx
```diff
@@ -17,6 +17,8 @@ const ButtonExample = () => {
       'chevron',
       'chevronRight',
       'close',
```
chriddell, line 22: Worth renaming `close` to avoid initial confusion?

chriddell: Sorry, maybe I spoke too soon. I can see we have:

- close
- closeBold
- closeMedium
- closeSmall
- closeCircle

Without looking at the variants, do these make logical sense? close, closeSmall and closeMedium read particularly unhelpfully to me. I would wonder why there is a `close` and `closeMedium` (they seem like they should be the same)

creido-welly: `closeMedium` is used in SiteHeader and will ultimately replace `close` (Header). As a temp solution it could be renamed (e.g. closeTemp) or commented to flag that this name isn't staying.

I agree there is some logic which needs addressing in the naming of some icons. There is also the larger task of deciding on how we render icons in general e.g. whether to have separate icons for each direction an arrow points or one icon which has a transform applied to it.

chriddell: Ah okay, so `close` is essentially deprecated?

That makes things simpler, if we added a comment to mention this in the file I think that would suffice?

creido-welly: Deprecation notice added.

I've gone a step further as there were a few broken icons which snuck in somewhere. They weren't being used so I've culled them to keep things uncluttered. This storybook entry also has a few additions.

chriddell (approved):
Looks great, very tidy

creido-welly:
> Looks great, very tidy

cheers

PR #451, opened 2021-03-22
8371 task: disabled styles

skibinska:
Related https://github.com/wellcometrust/corporate/issues/8371

### This PR

- implements the disabled styles: https://app.zeplin.io/project/5b4e06b48ae4580d4871178a/screen/6049f242e399d49556a502a4

### Example

Screenshot 2021-03-22 at 17 03 16

### DIFF
--- a/src/components/DateInput/DateInput.tsx
+++ b/src/components/DateInput/DateInput.tsx
@@ -4,6 +4,7 @@ import cx from 'classnames';
 type DateInputProps = {
   className?: string;
   describedBy?: string;
+  disabled?: boolean;
   id: string;
   isInvalid?: boolean;
   isRequired?: boolean;
@@ -12,7 +13,15 @@ type DateInputProps = {
 
 export const DateInput = forwardRef(
   (
-    { className, describedBy, id, isInvalid, isRequired, name }: DateInputProps,
+    {
+      className,
+      describedBy,
+      disabled,
+      id,
+      isInvalid,
+      isRequired,
+      name
+    }: DateInputProps,
     ref: React.Ref<HTMLInputElement>
   ) => {
     const classNames = cx('cc-date-input', {
@@ -24,6 +33,7 @@ export const DateInput = forwardRef(
       <input
         aria-describedby={describedBy}
         className={classNames}
+        disabled={disabled}
         id={id}
         name={name}
         ref={ref}
--- a/src/components/DateInput/DateInput.tsx
+++ b/src/components/DateInput/DateInput.tsx
@@ -4,8 +4,8 @@ import cx from 'classnames';
 type DateInputProps = {
   className?: string;
   describedBy?: string;
-  disabled?: boolean;
   id: string;
+  isDisabled?: boolean;
   isInvalid?: boolean;
   isRequired?: boolean;
   name: string;
@@ -16,8 +16,8 @@ export const DateInput = forwardRef(
     {
       className,
       describedBy,
-      disabled,
       id,
+      isDisabled,
       isInvalid,
       isRequired,
       name
@@ -33,7 +33,7 @@ export const DateInput = forwardRef(
       <input
         aria-describedby={describedBy}
         className={classNames}
-        disabled={disabled}
+        disabled={isDisabled}
         id={id}
         name={name}
         ref={ref}
--- a/src/components/Label/Label.tsx
+++ b/src/components/Label/Label.tsx
@@ -15,7 +15,7 @@ export const FormLabel = ({
   text
 }: LabelProps) => {
   const classNames = cx('cc-label', {
-    'cc-label--disabled': isDisabled,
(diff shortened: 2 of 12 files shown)


src/components/Label/Label.tsx
```diff
@@ -4,11 +4,18 @@ import cx from 'classnames';
 type LabelProps = {
   className?: string;
   htmlFor: string;
+  isDisabled?: boolean;
```
creido-welly, line 7: `disabled` is being used on everything else in this PR. Is there a specific reason you're using `isDisabled` here?

skibinska: I noticed that we are using `disabled` as a prop for [button](https://github.com/wellcometrust/corporate-components/blob/develop/src/components/Button/Button.tsx#L16) and [checkboxes](https://github.com/wellcometrust/funding-platform-frontend/blob/develop/src/components/CheckboxGroup/CheckboxGroup.jsx#L15), so for the tags that have a disabled attribute.

Label doesn't have a disabled attribute so I thought the way to differentiate it is to use `isDisabled` as it only adds a class.

Now when I am looking at the input component, we have [isRequired](https://github.com/wellcometrust/corporate-components/blob/develop/src/components/TextInput/TextInput.tsx#L9) prop, but maybe it should be `required`??

creido-welly: That logic makes sense, I thought that might be the case.

> Now when I am looking at the input component, we have isRequired prop, but maybe it should be required??

I agree, if you've taken that approach with `disabled` you should follow the same pattern for `required`

skibinska: ok, I will create a separate ticket to update it (I will have to make changes in CR too) - https://github.com/wellcometrust/corporate/issues/8377

src/components/DateInput/DateInput.tsx
```diff
@@ -12,7 +13,15 @@ type DateInputProps = {
 
 export const DateInput = forwardRef(
   (
-    { className, describedBy, id, isInvalid, isRequired, name }: DateInputProps,
+    {
+      className,
+      describedBy,
+      disabled,
```
chriddell: Could we use `isDisabled`? I think that is more consistent with the prop names of the other booleans.

src/components/Label/Label.tsx
```diff
@@ -4,11 +4,18 @@ import cx from 'classnames';
...
 
-export const FormLabel = ({ className, htmlFor, text }: LabelProps) => {
+export const FormLabel = ({
+  className,
+  htmlFor,
+  isDisabled,
+  text
+}: LabelProps) => {
   const classNames = cx('cc-label', {
+    'cc-label--disabled': isDisabled,
```
chriddell: I think our convention for "state" selectors is to use a new selector, e.g. `.is-disabled` (rather than `cc-label--disabled`).

@creido-welly to confirm.

src/components/RadioInput/RadioInput.tsx
```diff
@@ -6,6 +6,7 @@ import Label from 'Label';
 type RadioInputProps = {
   className?: string;
   describedBy?: string;
+  disabled?: boolean;
```
chriddell: Could we use `isDisabled` as the prop name (to follow boolean prop name convention)

src/components/TextArea/TextArea.tsx
```diff
@@ -4,6 +4,7 @@ import cx from 'classnames';
 type TextAreaProps = {
   className?: string;
   describedBy?: string;
+  disabled?: boolean;
```
chriddell: Same comment (could we use `isDisabled`)

src/components/TextInput/TextInput.tsx
```diff
@@ -4,6 +4,7 @@ import cx from 'classnames';
 type TextInputProps = {
   className?: string;
   describedBy?: string;
+  disabled?: boolean;
```
chriddell: And again (sorry, just making sure I catch them all!)... could we use `isDisabled`

chriddell (commented):
Left a few comments re: prop names - sorry, I'm always commenting on naming!

chriddell:
Ahh @skibinska - I'm sorry, I have just caught up with the comments between you and @creido-welly and can see that you guys agreed to change the props to `required` and `disabled`.

Personally I think they should be `isRequired` and `isDisabled` as that is the convention we have established for boolean props. ~However let's not drag this out, let's continue with what you've already done in this PR and I will delete my comments.~

~If we need to change these later, I will update the code.~

Actually I think it is worth having this conversation now... I'll post a message on Slack.

skibinska:
@creido-welly @chriddell this is ready to review again (I will create another PR for updating the button props)

chriddell (approved):
ace, looks good - sorry that you had to do this twice, but I think we have arrived at a better solution

PR #454, opened 2021-03-24
8388 task: change button prop name to isDisabled

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8388

### This PR

- changes the `disabled` prop name to `isDisabled` for `Button` component

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -74,7 +74,7 @@ export const Button = forwardRef(
         to={href}
         id={id}
         onClick={(e: ReactMouseEvent) => {
-          if (onClick && !isDisabled) {
+          if (typeof onClick === 'function' && !isDisabled) {
             onClick(e);
           }
         }}
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -13,12 +13,12 @@ type ButtonProps = {
   autoFocus?: boolean;
   className?: string;
   children: React.ReactNode;
-  disabled?: boolean;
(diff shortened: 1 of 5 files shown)


src/components/Button/Button.tsx
```diff
@@ -70,11 +70,11 @@ export const Button = forwardRef(
       <Element
         autoFocus={autoFocus}
         className={classNames}
-        disabled={disabled}
+        disabled={isDisabled}
         to={href}
         id={id}
         onClick={(e: ReactMouseEvent) => {
-          if (onClick && !disabled) {
+          if (onClick && !isDisabled) {
```
chriddell: I know this isn't part of your PR, but as you've already changed this line could you check the `typeof onClick === 'function'`, rather than checking that it is truthy please? 

That will guard better against (e.g.) someone providing the `onClick` prop as a string.

chriddell (approved):
Thanks for updating

src/components/NewsletterForm/NewsletterFormSubmit.tsx
```diff
@@ -18,7 +18,7 @@ export const NewsletterFormSubmit = ({
   <NewsletterFormItem type="submit">
     <Button
       className="newsletter-form__btn-submit"
-      disabled={disabled}
+      isDisabled={disabled}
```
chriddell, line 21: I have noticed we need to change the prop name in the NewsletterFormSubmit component too - but don't worry about this, I think we will move this component to corporate-react soon.

PR #455, opened 2021-03-25
8363 task: add input date fallback for Safari

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8363

### Context

1. `<input type="date">` is not supported in Safari. 

- unsupporting browsers gracefully degrade to a text input
- with date input supported, the value is normalized to the format `yyyy-mm-dd`, but with a text input, the browser has no recognition of what format the date should be in

One way around this is the `pattern` attribute on your date input. Even though the date picker doesn't use it, the text input fallback will.

see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

2. Instead of using `<input type="number">` we should use `<input type="text" inputmode="numeric" pattern="[0-9]*">` (this is following the gov.uk guidance)

see: https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/

### This PR

- adds `pattern` prop to `DateInput` and `TextInput`
- adds `inputMode` to `TextInput`

### DIFF
--- a/src/components/DateInput/DateInput.tsx
+++ b/src/components/DateInput/DateInput.tsx
@@ -9,6 +9,7 @@ type DateInputProps = {
   isInvalid?: boolean;
   isRequired?: boolean;
   name: string;
+  pattern?: string;
 };
 
 export const DateInput = forwardRef(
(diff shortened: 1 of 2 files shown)


creido-welly (approved):
New knowledge every day 🚀

PR #457, opened 2021-03-25
8371 task: add props and style for legend

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8371

### Context

To group a group of inputs (e.g radio buttons) correctly and give more context for screen reader users, we should wrap that group in a fieldset with legend.

Our design requires changing the colour of a label when input is disabled.

### This PR

- adds `isDisabled` and `legendClassName` props to legend

### DIFF
--- a/src/components/FormFieldset/FormFieldset.tsx
+++ b/src/components/FormFieldset/FormFieldset.tsx
@@ -16,18 +16,19 @@ export const FormFieldset = ({
   legend,
   legendClassName
 }: FormFieldProps) => {
-  const classNames = cx('cc-form-fieldset', {
-    [className]: className
-  });
-
-  const legendClassNames = cx('cc-form-legend', {
-    'is-disabled': isDisabled,
-    [legendClassName]: legendClassName
-  });
+  const classNames = {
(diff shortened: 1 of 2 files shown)


src/components/FormFieldset/FormFieldset.tsx
```diff
@@ -4,21 +4,30 @@ import cx from 'classnames';
...
 
 export const FormFieldset = ({
   children,
   className,
-  legend
+  isDisabled,
+  legend,
+  legendClassName
 }: FormFieldProps) => {
   const classNames = cx('cc-form-fieldset', {
```
creido-welly: There has been a recent move to having `classNames` as an object for readability, if there are multiple uses made of "classname" util, so this would become something along the lines of ...

```
const classNames = {
  fieldset: cx('cc-form-fieldset', {
    [className]: className
  }),
  legend: cx('cc-form-legend', {
    'is-disabled': isDisabled,
    [legendClassName]: legendClassName
  })
};
```

skibinska: updated :)

creido-welly (approved):

PR #463, opened 2021-04-07
8430 task: add hidden error text

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8430

### Context

The [gov.uk](https://design-system.service.gov.uk/components/error-message/) recommendation:

> To help screen reader users, the error message component includes a hidden ‘Error:’ before the error message. These users will hear, for example, “Error: The date your passport was issued must be in the past”.

### This PR

- adds the `VisuallyHidden` component
- adds visually hidden error text to `FormFieldErrror`

Screenshot 2021-04-07 at 11 17 40

### DIFF
--- a/src/components/VisuallyHidden/VisuallyHidden.tsx
+++ b/src/components/VisuallyHidden/VisuallyHidden.tsx
@@ -1,11 +1,22 @@
 import React from 'react';
 
 type VisuallyHiddenProps = {
-  text: string;
+  children: JSX.Element | JSX.Element[];
+  wrapperAs?: 'span' | 'div';
 };
 
-export const VisuallyHidden = ({ text }: VisuallyHiddenProps) => {
-  return <span className="u-visually-hidden">{text}&nbsp;</span>;
+export const VisuallyHidden = ({
+  children,
+  wrapperAs = 'span',
+  ...props
+}: VisuallyHiddenProps) => {
+  const WrapperElement = wrapperAs;
+
+  return (
+    <WrapperElement className="u-visually-hidden" {...props}>
+      {children}
+    </WrapperElement>
+  );
 };
 
 export default VisuallyHidden;
--- a/src/components/VisuallyHidden/VisuallyHidden.tsx
+++ b/src/components/VisuallyHidden/VisuallyHidden.tsx
@@ -1,0 +1,11 @@
+import React from 'react';
+
+type VisuallyHiddenProps = {
+  text: string;
+};
+
+export const VisuallyHidden = ({ text }: VisuallyHiddenProps) => {
+  return <span className="u-visually-hidden">{text}&nbsp;</span>;
+};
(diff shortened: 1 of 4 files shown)


src/components/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+type VisuallyHiddenProps = {
+  text: string;
+};
+
+export const VisuallyHidden = ({ text }: VisuallyHiddenProps) => {
```
creido-welly: @chriddell what are your thoughts on this?

I like the accessibility but I'm not sure if it's worth having a React component for this. I'm not completely opposed though.

If we decide it is worth having this component then the space should be added in the `text` prop and not in the component since it might be used elsewhere in a sentence.

chriddell: I think it could be useful to have a component for this because...

- easier to change CSS selector (e.g. `.u-visually-hidden` could become `.u-hide-visually`/`.is-hidden` in a one line change)
- approach to hiding content becomes explicitly consistent

One thing I would suggest is using a `children` prop rather than `text` prop, to allow the component to render either other React components or HTML elements, e.g.

```
<VisuallyHidden wrapperAs="div">
  <h2>Title of the long text</h2>
  <p>This is a longer piece of text which is hidden</p>
</VisuallyHidden>
```

There may be scenarios where we want to add aria-* attributes to the component; I think adding these props to the component is something worth considering.

skibinska: I have updated `VisuallyHidden` component :)

src/components/VisuallyHidden/VisuallyHidden.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+type VisuallyHiddenProps = {
+  children: JSX.Element | JSX.Element[];
+  wrapperAs?: 'span' | 'div';
+};
```
chriddell: I think we can declare the type using the `HTMLAttributes` type from React...

```ts
import React, { HTMLAttributes } from 'react';

type VisuallyHiddenProps = (
  | HTMLAttributes<HTMLDivElement>
  | HTMLAttributes<HTMLSpanElement>
) & {
  children: JSX.Element | JSX.Element[] | HTMLElement;
  wrapperAs?: 'span' | 'div';
};
```

What this does, I think, is guard against random props being passed to VisuallyHidden which do not exist on HTMLAttributes.

If I add the above code, and then hover over `...props` in VSCode, I see:

Screenshot 2021-04-08 at 16 08 35

Which helps understand what props can be passed to VisuallyHidden.

chriddell (approved):
Looks good to me, thanks for updating

PR #465, opened 2021-04-08
8440 task: add accessible attributes to ResultsCount component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8440

### Context

Where dynamic content is updated either automatically or in response to a user action, we need to make sure that users are notified of this update in an accessible way.

Our listing pages have filters/ a search bar. Selecting a checkbox will cause the search results to update. This style of interaction presents a problem for users of screen readers because it relies on the user being able to **see** the change in order to know it happened.

The solution to this is to use an ARIA live region. Content within an ARIA live region will be announced when it changes, so we can wrap our updating content and as soon as the user does anything to affect the results being shown they will be notified.

### This PR

- adds aria-live and aria-atomic to the `ResultsCount` component

### DIFF
--- a/src/components/ResultsCount/ResultsCount.tsx
+++ b/src/components/ResultsCount/ResultsCount.tsx
@@ -32,7 +32,7 @@ export const ResultsCount = ({
 
   return (
     <div className={classNames}>
-      <p className="cc-results-count__result">
+      <p className="cc-results-count__result" aria-live="polite" role="alert">
         Showing{' '}
         <strong>
           {currentCount} results of {resultsCount}
--- a/src/components/ResultsCount/ResultsCount.tsx
+++ b/src/components/ResultsCount/ResultsCount.tsx
@@ -32,7 +32,11 @@ export const ResultsCount = ({
 
   return (
     <div className={classNames}>
-      <p className="cc-results-count__result" aria-live="polite" role="alert">
+      <p
+        className="cc-results-count__result"
+        aria-atomic="true"
+        aria-live="polite"
+      >
         Showing{' '}
         <strong>
           {currentCount} results of {resultsCount}


src/components/ResultsCount/ResultsCount.tsx
```diff
@@ -32,7 +32,7 @@ export const ResultsCount = ({
 
   return (
     <div className={classNames}>
-      <p className="cc-results-count__result">
+      <p className="cc-results-count__result" aria-live="polite" role="alert">
```
chriddell: Is the use of `role="alert"` here appropriate? From [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role):

> Because of its intrusive nature, the alert role must be used sparingly and **only in situations where the user's immediate attention is required.** Dynamic changes that are less urgent should use a less aggressive method, such as aria-live="polite" or other live region roles.

creido-welly: As I understand it, adding `role="alert"` implies that an element will be assertive (the same as adding `aria-live="assertive"` to the element). If `aria-live="polite"` is used the screen reader will wait until after it has finished its current thread before announcing the alert.

creido-welly: ☝️  it was in session one of the a11y training yesterday :)

creido-welly (approved):

src/components/ResultsCount/ResultsCount.tsx
chriddell: Thanks Chris. re: assertive, I can see that's mentioned in the notes from the MDN docs: 

> Using the alert role on an element implies that element has aria-live="assertive".

My concern is whether aria-live="polite" and role="alert" are conflicting. There's no mention of using the combination in the MDN docs, but there is this line:

> Dynamic changes that are less urgent should use a less aggressive method, such as aria-live="polite" or other live region roles. 

Could either of you suggest any other docs I could take a look at regarding aria-live & role="alert"?

skibinska:
@chriddell, I think you are correct. I shouldn't be using the `role="alert"` in this context.

What I was trying to achieve was to announce to a screen reader user the whole sentence, but the message I was getting was `1-20 42` (only with `aria-live`). 

Screenshot 2021-04-15 at 17 44 56

So, I 'forced' it to read this text by using `role alert` .

What I didn't get was that 
> Assistive technologies will **announce dynamic changes** in the content of a live region.

So in my case the dynamic changes are numbers not the whole text.

After checking the slides from the accessibility training the `aria-atomic` seems to be a better option.

I tested it and it works as expected.

Screenshot 2021-04-15 at 18 24 13

creido-welly (approved):

PR #467, opened 2021-04-09
8438 task: replaces u-visually-hidden with component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8438

### This PR

Replaces all `u-visually-hidden` classes with `VisuallyHidden` component.

---

### QUESTIONS

What should we do about components where `u-visually-hidden` is either passed as a className prop or `hideText` adds visually hidden class?

Examples:

1. [SearchFormLabel](https://github.com/wellcometrust/corporate-components/blob/develop/src/components/SearchForm/SearchFormLabel.tsx#L19)

My suggestion:

```javascript
export const SearchFormLabel = ({
  hideText, ...
}) => {
  const classNames = cx({
    'search-form__label': !className,
    'u-visually-hidden': hideText,<-- remove this line 
    [`${className}`]: className
  });

return hideText ? (
    <VisuallyHidden>
      <label className={classNames} htmlFor={htmlFor}>
        Search
      </label>
    </VisuallyHidden>
  ) : (
    <label className={classNames} htmlFor={htmlFor}>
      Search
    </label>
  );

};
```

2. Button in [SidebarSearch](https://github.com/wellcometrust/corporate-components/blob/develop/src/components/SidebarSearch/SidebarSearch.tsx#L57)

```javascript
<Button textClassName="u-visually-hidden">Search</Button>
```
My suggestion: Instead of using the class we could use Visually Hidden

```javascript
<Button><VisuallyHidden>Search</VisuallyHidden></Button>
```

3. Should we add a string to the `VisuallyHiddenProps` to have: `children: JSX.Element | JSX.Element[] | HTMLElement | string;`

So we could use it like this:
```html
<VisuallyHidden>Search</VisuallyHidden>
```

instead of

```html
<VisuallyHidden><>Search</></VisuallyHidden>
```


src/components/Pagination/Pagination.tsx
```diff
@@ -20,7 +21,7 @@ export const PaginationButtonNext = () => (
 
 export const PaginationButtonBreak = () => (
   <>
-    <span className="u-visually-hidden">go to the next page group&nbsp;</span>…
```
creido-welly, line 23: Minor: Is the React Fragment still needed?

creido-welly (approved):

src/components/Pagination/Pagination.tsx
skibinska: I am getting a parsing error without it. I think we need a wrapper for visually hidden text and `…`

PR #469, opened 2021-04-12
8439 task: add Remove filter [filter] visually hidden text

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8439

### Context

This is updated to follow the pattern seen on [the gov.uk page](https://www.gov.uk/search/all?keywords=tenant&level_one_taxon=3cf97f69-84de-41ae-bc7b-7e2cc238fa58&order=relevance)

This is read: `Remove filter Environment`
Screenshot 2021-04-12 at 11 23 23

### This PR

- adds visually hidden `Remove filter` to `SidebarFilter`

### DIFF
--- a/src/components/SidebarFilter/SidebarFilter.tsx
+++ b/src/components/SidebarFilter/SidebarFilter.tsx
@@ -4,6 +4,7 @@ import cx from 'classnames';
 import Button from 'Button';
 import Accordion, { AccordionItem } from 'Accordion/Accordion';
 import Checkbox from 'Checkbox';
+import VisuallyHidden from 'VisuallyHidden';
 
 type SidebarFilterProps = {
   activeTags: {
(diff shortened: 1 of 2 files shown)


chriddell (approved):

src/components/SidebarFilter/SidebarFilter.tsx
```diff
@@ -71,6 +72,7 @@ export const SidebarFilter = ({
               onClick={() => onTagRemove(value)}
               type="button"
             >
+              <VisuallyHidden>Remove filter&nbsp;</VisuallyHidden>
```
chriddell: What is the reason for the non-breaking space?

skibinska: I wanted to make sure there is a space between `Remove filter` and the filter name. So, to have the sentence like `Remove filter&nbsp;Biomedical science research`. Should I use space instead?

chriddell: Ah okay, I think a "normal" space is more appropriate. `&nbsp;` is used for visual reasons (when you don't want a word to break on to a new line). You may need to use a template literal so that the linter doesn't remove the space, something like:

```
{`Remove filter `}
```

or

```
Remove filter{` `}
```

PR #470, opened 2021-04-12
8436 task: add SkipLink component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8436

### Context

[gov.uk](https://www.gov.uk/search/all?keywords=tax) page has a `Skip to results` link above the filter section

### This PR

- adds `SkipLink` component

Screenshot 2021-04-12 at 17 18 49

### Test

- fetch this branch and checkout 
- run `npm run storybook`

### DIFF
--- a/src/components/SidebarFilter/SidebarFilter.tsx
+++ b/src/components/SidebarFilter/SidebarFilter.tsx
@@ -47,7 +47,7 @@ export const SidebarFilter = ({
 
   return (
     <div className={classNames}>
-      {skipLink && <SkipLink href={skipLink} skipLinkText="Skip to results" />}
+      <SkipLink href={skipLink} text="Skip to results" />
       <header className="cc-sidebar-filter__header">
         <h2 className="cc-sidebar-filter__header-title">Refine results</h2>
         <span className="cc-sidebar-filter__header-meta">
--- a/src/components/SidebarFilter/SidebarFilter.tsx
+++ b/src/components/SidebarFilter/SidebarFilter.tsx
@@ -18,7 +18,7 @@ type SidebarFilterProps = {
   onChange: (value: string) => void;
   onClear: () => void;
   onTagRemove: (value: string) => void;
-  skipLink: string;
+  skipLinkHref: string;
   tags: {
     name: string;
     items: {
@@ -38,7 +38,7 @@ export const SidebarFilter = ({
   onChange,
   onClear,
   onTagRemove,
-  skipLink,
+  skipLinkHref,
   tags
 }: SidebarFilterProps) => {
   const classNames = cx('cc-sidebar-filter', {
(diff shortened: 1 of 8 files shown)


chriddell (commented):
I think this is a great idea, but have one concern about the approach and worry this is slightly over-engineered...

Could you explain your reasons for writing this as a new component?

My thinking is: if this is only used in the SidebarFilter component we could write the markup directly inside that component rather than creating a new component (SkipLink) and exporting it.

I'm more than happy to change my mind if we think this is going to be useful in places other than the SidebarFilter.

skibinska:
I am using this component to add `Skip to main content` link in the Layout component and on the grants awarded search page (we are not using SidebarFilter on that page).

[CR PR](https://github.com/wellcometrust/corporate-react/pull/924)

chriddell:
@skibinska Thanks for that. I agree this would benefit from being a new component, then!

src/components/SkipLink/SkipLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+  href: string;
+  isNarrow?: boolean;
+  skipLinkText: string;
+};
+
+export const SkipLink = ({
+  className,
+  href,
+  isNarrow,
+  skipLinkText
```
chriddell: As we are already inside the SkipLink component, would it be more concise/appropriate to name this prop `text`? It might read better when writing:

```
<SkipLink {...props} text="Skip to content" />
```

src/components/SidebarFilter/SidebarFilter.tsx
```diff
@@ -16,6 +17,7 @@ type SidebarFilterProps = {
   onChange: (value: string) => void;
   onClear: () => void;
   onTagRemove: (value: string) => void;
+  skipLink: string;
```
chriddell: Should this be a boolean prop?

src/components/SkipLink/SkipLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+}: SkipLinkProps) => {
+  const classNames = {
+    wrapper: cx('cc-skip-link', {
+      [className]: className
+    }),
+    link: cx('cc-skip-link__link', {
+      'cc-skip-link__link--narrow': isNarrow
```
creido-welly: `isNarrow` is a little confusing as a property name. If this controls how the link is positioned (absolute or static) how about something more in line with that and maybe even make it enumerable for ease of expansion in future?

type
```
position: 'static' | 'float';
```

prop
```
position = static
```

classes
```
const classNames = {
    wrapper: cx('cc-skip-link', {
      [className]: className
    }),
    link: cx('cc-skip-link__link', {
      'cc-skip-link__link--float': position === 'float'
  })
};
```

src/components/SidebarFilter/SidebarFilter.tsx
skibinska: I am passing here an `href` for SkipLink, so:

```javascript
export const SidebarFilter = ({
  skipLink,
  ...
}) => {
  return (
    <div className={classNames}>
      <SkipLink href={skipLink} text="Skip to results" />
      <header className="cc-sidebar-filter__header">...</header>
    </div>
```

and then
```javascript
<SidebarFilter skipLink="#results" ...props/>
```

creido-welly (approved):

src/components/SidebarFilter/SidebarFilter.tsx
chriddell: Ahh okay, how about `skipLinkHref` to make it slightly clearer?

PR #471, opened 2021-04-14
8450 task: add autocomplete and spellcheck attributes to TextInput

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8450

### Context

> Use the autocomplete attribute on the text input component when you’re asking for a user’s name. This lets browsers autofill the information on a user’s behalf if they’ve entered it previously.

> Sometimes, browsers will spellcheck the information a user enters into a text input. To make sure user’s names will not be spellchecked, set the spellcheck attribute to false

source: [gov.uk design system: names pattern](https://design-system.service.gov.uk/patterns/names/)

After accessibility training and checking the gov.uk design system we should:

- set the autocomplete attribute to `name` - if asking for a user’s full name in a single field
- set the autocomplete attribute `tel` on telephone number inputs
- set the autocomplete attribute to `email` on email inputs
- do not spellcheck user’s names nor email addresses

**sources**

- [gov.uk design system: names pattern](https://design-system.service.gov.uk/patterns/names/)
- [gov.uk design system: telephone numbers pattern](https://design-system.service.gov.uk/patterns/telephone-numbers/)
- [gov.uk design system: email pattern](https://design-system.service.gov.uk/patterns/email-addresses/)

### This PR

- adds `autoComplete` and `spellCheck` props to `TextInput`

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -2,9 +2,10 @@ import React, { forwardRef } from 'react';
 import cx from 'classnames';
 
 type TextInputProps = {
-  autoComplete?: 'name' | 'tel' | 'email';
+  autoComplete?: string;
   className?: string;
   describedBy?: string;
+  hasSpellCheck?: boolean;
   id: string;
   inputMode?: 'numeric';
   isDisabled?: boolean;
@@ -12,7 +13,6 @@ type TextInputProps = {
   isRequired?: boolean;
   name: string;
   pattern?: string;
-  spellCheck?: 'true' | 'false';
   type: 'text' | 'email' | 'tel';
 };
 
@@ -22,6 +22,7 @@ export const TextInput = forwardRef(
       autoComplete,
       className,
       describedBy,
+      hasSpellCheck,
       id,
       inputMode,
       isDisabled,
@@ -29,7 +30,6 @@ export const TextInput = forwardRef(
       isRequired,
       name,
       pattern,
-      spellCheck,
       type = 'text'
     }: TextInputProps,
     ref: React.Ref<HTMLInputElement>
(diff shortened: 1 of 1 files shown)


src/components/TextInput/TextInput.tsx
```diff
@@ -11,12 +12,14 @@ type TextInputProps = {
   isRequired?: boolean;
   name: string;
   pattern?: string;
+  spellCheck?: 'true' | 'false';
```
chriddell: Would a boolean prop be more appropriate here? If necessary, we could convert it to a string in the render method: `hasSpellCheck.toString()`

src/components/TextInput/TextInput.tsx
```diff
@@ -2,6 +2,7 @@ import React, { forwardRef } from 'react';
 import cx from 'classnames';
 
 type TextInputProps = {
+  autoComplete?: 'name' | 'tel' | 'email';
```
chriddell: Is it worth adding more values here? Looking at MDN, there are many possible values for autocomplete: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete.

I did some testing and I think this works:

```
autoComplete?: HTMLInputElement['autocomplete'];
```

chriddell: Ahh, that only declares it as a string. I was hoping it would be an exhaustive list of all of the values available for `autocomplete`.

Still, maybe better to declare the type as a string? That way we will not need to extend the component if we want to add more values for `autocomplete` in the future.

src/components/TextInput/TextInput.tsx
skibinska: yeap, I will update this :)

skibinska:
this is ready to review again :)

src/components/TextInput/TextInput.tsx
creido-welly: hasSpellCheck also follows the convention we've established with other form control properties e.g. `isRequired` `isValid` etc

creido-welly (approved):

PR #472, opened 2021-04-19
8450 task: sets spellCheck prop as enumerable

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8450

### Context

The spellcheck attribute is an enumerated one and not a Boolean.
This means that the explicit usage of one of the values true or false is mandatory.
When the spellcheck attribute is not included, the default value depends on the element and browser.
 
We are following the gov.uk pattern to add spellcheck="false" on name and email fields.

### This PR

- implements @creido-welly 's comment on setting `spellCheck` as enumerable. Comment can be seen here: https://github.com/wellcometrust/corporate-react/pull/930#discussion_r615718086

### DIFF
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -3,7 +3,7 @@ import cx from 'classnames';
 
 /**
  * The spellcheck attribute is an enumerated one and not a Boolean.
- * This means that the explicit usage of one of the values true or false is mandatory.
+ * This means that only one of the string values 'true' or 'false' may be used.
  * When the spellcheck attribute is not included, the default value depends on the element and browser.
  *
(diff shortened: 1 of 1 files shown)


src/components/TextInput/TextInput.tsx
```diff
@@ -1,18 +1,29 @@
 import React, { forwardRef } from 'react';
 import cx from 'classnames';
 
+/**
+ * The spellcheck attribute is an enumerated one and not a Boolean.
+ * This means that the explicit usage of one of the values true or false is mandatory.
```
creido-welly: This reads as if you intend this to be a mandatory prop. Probably better written as "This means that only one of the string values 'true' or 'false' may be used."

creido-welly (approved):
One minor comment re your comment wording 🚓   but otherwise good to go

PR #474, opened 2021-04-21
8447 feature(style): change line-height to 1.6

skibinska:

### Context

The new design introduces only one line-height - **1.6**

I have created this branch as a feature and I am going to use it as a base for the rest of my work, so I can split this into smaller chunks.

I checked CR and couldn't find any useage of  `--body-line-height-md` and `--body-line-height-sm`.

### This PR

- removes redundant line heights variables
- adds one line-height variable for body line height -> `--body-line-height`


creido-welly (commented):
Code-wise this all looks fine.

Regarding the approach, I wonder if we need to declare line-height so widely if everything is based on a single value.

The consuming app (corporate-react) currently declares a global line-height of 1.7  (on the body tag) so it makes sense to apply overrides to individual components but if this is also changing (which it should) then we may not need them.

Moving forward, from a design system point of view, would we want to keep overrides, for example, to prevent a consuming app breaking convention?

What do you think?

creido-welly:
As discussed, line-height style rules for individual components will be kept for the time-being. Long-term usage rules for line-height will be established in the design system.

creido-welly (approved):

chriddell (approved):

PR #480, opened 2021-04-23
8447 task(style): updates letter spacing

skibinska:

### Context

We have 3 different letter-spacing values now:

- Body styles: **letter-spacing: 0.4px**;
- Heading styles:
   - font-family: Wellcome-Bold; **letter-spacing: -0.5px**;
   - font-family: HelveticaNeue-Medium; **letter-spacing: 0.2px**;

### This PR

- add new letter-spacing variables
- Jason wants to try the fixed units for the letter-spacing. We will iterate on it after QA

### DIFF
--- a/src/assets/styles/10-settings/_typography.scss
+++ b/src/assets/styles/10-settings/_typography.scss
@@ -42,8 +42,8 @@
 
   // letter spacing
   --body-letter-spacing: 0.4px;
-  --heading-bold-letter-spacing: -0.5px;
-  --heading-letter-spacing: 0.2px;
+  --heading-main-letter-spacing: -0.5px;
+  --heading-base-letter-spacing: 0.2px;
 
   @include mq(sm) {
     --heading-xl: calc(32/16 * var(--text-base-size));
--- a/src/assets/styles/10-settings/_typography.scss
+++ b/src/assets/styles/10-settings/_typography.scss
@@ -26,7 +26,7 @@
   --body-md: calc(14/16 * var(--text-base-size));     // Small
   --body-sm: calc(13/16 * var(--text-base-size));     // Smaller
   --body-xs: calc(11/16 * var(--text-base-size));     // Smallest
-  --body-xxs: calc(10/16 * var(--text-base-size));    // 
+  --body-xxs: calc(10/16 * var(--text-base-size));    //
 
   // font size responsive headings
   --heading-xxl: calc(28/16 * var(--text-base-size)); // Super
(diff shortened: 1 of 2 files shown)


src/assets/styles/10-settings/_typography.scss
```diff
@@ -41,8 +41,9 @@
   --heading-line-height: var(--body-line-height);
 
   // letter spacing
-  --body-letter-spacing: 0.03125rem;
-  --heading-letter-spacing: -0.014em;
+  --body-letter-spacing: 0.4px;
+  --heading-bold-letter-spacing: -0.5px;
```
creido-welly: What is your reasoning behind using "bold" in the name?

skibinska: We are using this variable only for headings with Wellcome Bold font. Wanted to make it clear that we are using it for this reason...

skibinska: Happy to change it!

creido-welly: The word "bold" is tied in too much with heading style and especially as they are now (using Wellcome-Bold). I would've suggested `--heading-primary-letter-spacing` but when it comes to font families Wellcome-Bold is actually set as `--font-secondary` so that seems odd.

@chriddell any thoughts?

chriddell: Hmm, I agree that using "bold" ties us to a "bold" font (which may change) and is also confusing as our font is literally named "Bold", too....

I can see we're using 3 letter-spacing values, so how about we tie their names to the scenarios they are used in? Taking the "Wellcome Bold" font as an example - what is the scenario(s) in which we use "Wellcome Bold"? e.g. is it all h1s and h2s? Is it when they are used in certain components? Is it when a heading is above a certain font-size?

Presumably there is a rule for when to use Wellcome Bold, and when to use Helvetica. If we know the rule, we can name our letter-spacing accordingly.

Would ideally end up something like

```
--body-letter-spacing: xx;
--heading-large-letter-spacing: xx; // equivalent to --heading-bold-letter-spacing
--heading-small-letter-spacing: xx;
```

The latter two variable names should be changed depending on what the design rule is for using Wellcome Bold or Helvetica.

skibinska: My understanding is that only that theoretical `<h0>` and `<h1>` are using Wellcome Bold and only these two headings have letter-spacing **-0.5px**.

creido-welly (approved):

chriddell (approved):

PR #486, opened 2021-05-06
8447 task(style): design QA

skibinska:

### Context

### This PR

Implements Jason's design QA. More can be seen here: https://docs.google.com/spreadsheets/d/1uUWk7s8b_nr6F2upj06RShHQPW2dgB2N0zyRyYioFR4/edit#gid=0

- removes custom headings sizes within accordion
- adds font-weight 500 to Helvetica headings
- update space for `FeaturedPromo` and `ImageCard` components
- removes fancy underline from H3 in `RichText` nad `Timeline`

### DIFF
--- a/src/assets/styles/10-settings/_typography.scss
+++ b/src/assets/styles/10-settings/_typography.scss
@@ -47,7 +47,7 @@
   --heading-minor-letter-spacing: 0.2px;
 
   // font weight
-  --font-weight: 500;
+  --heading-font-weight: 500;
 
   @include mq(sm) {
     --heading-xl: calc(32/16 * var(--text-base-size));
--- a/src/assets/styles/10-settings/_typography.scss
+++ b/src/assets/styles/10-settings/_typography.scss
@@ -46,6 +46,9 @@
   --heading-major-letter-spacing: -0.5px;
   --heading-minor-letter-spacing: 0.2px;
 
+  // font weight
+  --font-weight: 500;
+
   @include mq(sm) {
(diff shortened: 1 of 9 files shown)


src/assets/styles/10-settings/_typography.scss
```diff
@@ -46,6 +46,9 @@
   --heading-major-letter-spacing: -0.5px;
   --heading-minor-letter-spacing: 0.2px;
 
+  // font weight
+  --font-weight: 500;
```
creido-welly: `--heading-font-weight` gives better context (if I have the context correctly?)

skibinska: I was thinking about similar variable name, but then realised we use this font-weight in various components, so wanted to make it more general.

creido-welly: Fair enough, but I think it should have more context in the name if it's not used globally. I know it's associated with Helvetica but can you list some of the contexts it is used in?

skibinska: yes, the font-weight: 500 is used on all Helvetica headings.

I found that it is used for:

- cc-page-title__meta-label,
- cc-inpage-nav__title,
- th,
- c-hero-statement__title.

Looking at it again, I think it would be better to rename it to `--heading-font-weight`

creido-welly: Yes I think you're right

HUZZAH!

creido-welly (approved):
👍

PR #487, opened 2021-05-10
task: update dependencies

skibinska:
### Context

We have 3 `Dependabot alerts`:

Screenshot 2021-05-10 at 11 45 17

### This PR

- adds updates after running `npm audit fix` and `npm audit fix --force`
- adds plugins configuration to `babel.config.js` to stops that warning coming up

**Example warning:**

> the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-property-in-object since the "loose" mode option was set to "true" for @babel/plugin-proposal-private-methods.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding ["@babel/plugin-proposal-private-property-in-object", { "loose": true }] to the "plugins" section of your Babel config.

**Current state**

We have `2 vulnerabilities (1 low, 1 moderate)`, which could be fixed by `npm audit fix`. However after running it I am getting a message: `6  moderate vulnerabilities`... so this change is not include in this PR.


babel.config.js
```diff
@@ -1,24 +1,31 @@
...
       '@babel/plugin-transform-runtime',
       {
-        'corejs': 3
+        corejs: 3
       }
-    ]
+    ],
+    /**
+     * Add these options to silence the Storybook build warnings about different loose mode configuration
+     * @see https://github.com/babel/babel/issues/11622
```
chriddell: I think this should follow the jsdoc formatting for a URL:

`@see {@link https://github.com/babel/babel/issues/11622}`

It's only a minor point, but it should help with consistency across the codebase if we follow jsdoc formatting exactly. 

Also (I think) following the `@link` convention means jsdoc add-ons for our editor software will be able to use syntax highlighting for these comments, to make them a little more readable!

chriddell (commented):
I notice this adds 35,000 lines to the package-lock.json... I'm wondering if that means the lockfileVersion has changed accidentally? Maybe `npm i` has been used from a newer version of node/npm than that which we want to work with?

It seems like a lot of new lines, when only two packages have been updated. But I can't be certain...

babel.config.js
skibinska: @chriddell, I run these commands on a different branch and this is what I was getting:

**npm audit fix:**
- updates only `package-lock.json`
- it changes `lockfileVersion` from 1 to 2 (if we don't want it I am not sure how to undo it)
- `added 314 packages, removed 148 packages, changed 177 packages, and audited 3186 packages in 49s`

**npm audit fix --force:**
- updates `package-lock.json` and `package.json`
- `added 36 packages, removed 89 packages, changed 26 packages, and audited 3133 packages in 12s`

creido-welly:
@chriddell I would also assume the massive diff for package-lock.json is due to a more up to date version of node which would then bring about a change in lockfileVersion from 1 to 2.

A while back we upgraded corporate-react to node 15.0.1 but it appears we haven't done similarly for CC. Can you foresee any issues updating? It would mean we can use the same node version for both.

chriddell:
Yeah I think (*hope...) it will be fine to update, and makes sense to do so now seeing as we've updated corp-react. I think best way to test would be to run corp-components locally and link it to corporate-react, then run corporate-react locally and try and spot any errors?

Off the top of my head, that's the only way we can test this? I'm happy to do that, so I'll give it a go now using this branch.

babel.config.js
chriddell: Thanks for the explainer, I've left a comment further below in response to Chris' about how we can test this. I think it's fine for the lockfile version to update, but I'll do some testing locally before we merge this branch to see if I can spot anything obviously broken.

Will comment again on this PR when I've done.

skibinska: thanks Chris!

skibinska:
@chriddell there will be some errors/warnings regarding missing variables, as I haven't merged [the heading update PR](https://github.com/wellcometrust/corporate-react/pull/948) in corporate-react.

chriddell:
I'm running in to lots of trouble with missing scripts when I pull this branch and run `npm run dev` (after running `npm ci` to install the packages). 

I've started looking in to this and I'm down a wormhole, but before I start to put a fix together I want to check - can either of you successfully install the dependencies for this branch and run `npm run dev`?

Could you try doing so and let me know how you get on?

The first error I hit when I run `npm ci && npm run dev` is (something like):

```
sh: tsc: command not found
npm ERR! code 127
npm ERR! path /www/sites/corporate-components
npm ERR! command failed
npm ERR! command sh -c npm run build:types && rollup -c ./config/rollup.config.js -w
```

(I can fix this one, but then more errors start to happen)

skibinska:
@chriddell, yes I am getting the same error. Sorry, I don't know how I missed that.

chriddell:
Thanks @skibinska - no worries, I just wanted to check it wasn't only my machine.

I think the simplest thing for now might be to try and fix the issues related to the dependabot alerts **without** updating the version of node/npm we are using for corporate-components.

I think if we are to update corp-components to use node v15.0.1 & npm v7.0.3 we're going to have to deal with a lot of issues from out-of-date dependencies. This is something that we will have to do at some point, but for now if we can fix the dependabot alerts without causing too much pain, that's preferred.

I'll try and do this today and see what happens. If it's not possible, we'll probably have to update a lot of packages from our package.json to their latest versions in order to use node v15.0.1 and npm 7.0.3.

PR #490, opened 2021-05-11
8536 task(chore): fix dependabot alerts

creido-welly:
Relates to wellcometrust/corporate#8536


chriddell (approved):
Happy for this to get merged if it seems there are no issues with the built files created as a result?

package.json
```diff
@@ -24,6 +24,7 @@
     "lint": "npm run lint:style && npm run lint:js",
     "lint:js": "eslint --fix ./src --ext .js,.jsx,.ts,.tsx && echo 'JS linting complete'",
     "lint:style": "stylelint --fix '**/*.scss' && echo 'Style linting complete'",
+    "preinstall": "npx npm-force-resolutions",
```
chriddell, line 27: Not fully sure what this does, but do you think it will be a problem when this builds on the server? (e.g. does the remote server have access to `npx` commands? does it need to?)

chriddell: *does this actually build on the server? Or do we basically just upload the built files as they are?

creido-welly: It's a very good question. I've done what I can to test it locally - Storybook works absolutely fine and `npm publish --dry-run` completed without any issues. I guess the real proof will be in how it deploys to github pages.

As as aside, I came across this package through a Google search after re-reading the thread about trim and `remark-parse` - https://github.com/storybookjs/storybook/issues/14603. It does indeed fix the vulnerabilities but flagged the following in the output.

Screenshot 2021-05-11 at 12 54 26

creido-welly: `preinstall` will run before npm install basically so it runs when we publish to npm. I expect the package to remain unaffected as storybook is a dev dependency and remark-parse will only be used [here](http://ghp.wellcomecollection.org/corporate-components).

chriddell (approved):

PR #493, opened 2021-05-14
8561 task: updates SiteAlert props

chriddell:
# Context

As part of the preview work being done over at /issues/7988, we want to use the SiteAlert (aka "Collection banner") style for the "Preview mode" message.

To do this, it's necessary to pass a custom classname to the SiteAlert (so that we can style it accordingly, when the SiteAlert exists in a different place in the DOM - e.g. not as a child of the site header).

We also don't want to pass the `handleDismiss` callback, because the user should not be able to close this message; it should always remain present. Therefore we want to make the `handleDismiss` prop optional.

# Description

This PR:

- sets `handleDismiss` optional
- conditionally renders the `<button>` element based on presence of `handleDismiss`
- adds `className` prop

### DIFF
--- a/src/components/SiteAlert/_site-alert.scss
+++ b/src/components/SiteAlert/_site-alert.scss
@@ -73,6 +73,9 @@
 
 .site-alert__text {
   @extend %margin-0;
+
+  text-align: center;
+  width: 100%;
 }
 
 .site-alert__btn-close {
@@ -92,9 +95,4 @@
     margin-left: var(--space-unit);
     width: calc(1.5 * var(--space-unit));
   }
-}
-
-.site-alert__text {
-  width: 100%;
-  text-align: center;
 }
--- a/src/components/SiteAlert/_site-alert.scss
+++ b/src/components/SiteAlert/_site-alert.scss
@@ -93,3 +93,8 @@
     width: calc(1.5 * var(--space-unit));
   }
 }
+
+.site-alert__text {
+  width: 100%;
+  text-align: center;
+}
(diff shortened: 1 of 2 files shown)


creido-welly (approved):
All good.

Do you see any reason why this shouldn't be moved to corporate-react? Not pushing for moving it in this PR, but if it can be moved you could add a jsdoc `@deprecated ...` notice and ticket ([see Footer for an example](https://github.com/wellcometrust/corporate-components/blob/dd48b6d474f8d863105f2d0af45672d385dbb625/src/components/Footer/Footer.tsx#L1)).

chriddell:
That's a good shout @creido-welly - I think there's reason for this to stay in corporate-components. It's a general alert style, which could be reasoned about either way, I think.

I think there's scope for using this in other consuming apps, but I have no strong feelings either way.

I'm going to merge this as is, mostly because I'd like to clear something off my desk before the end of the week! But I would be happy to deprecate and move this if we think it should be.

src/components/SiteAlert/_site-alert.scss
```diff
@@ -93,3 +93,8 @@
     width: calc(1.5 * var(--space-unit));
   }
 }
+
+.site-alert__text {
+  width: 100%;
```
creido-welly: Looks like there is already a rule in this file for `.site-alert__text`

chriddell: Ah, missed that. (Something I thought the linter would pick up... will look at that separately)

PR #494, opened 2021-05-17
8554 task: yellow links and buttons

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8554

Link to designs: https://www.sketch.com/s/fdfb45f3-3e86-49d2-9f06-332acc86024c

### My approach

I added new link styles directly to the components in CC, and I didn't add any global styles to anchor in CR.

### This PR

- adds new link and button styles
- adds `animated-link` mixin which can handle two scenarios
   1. the regular anchor
   2. the anchor with children (adds the underline styles to chosen child)
- adds `u-animated-link` - I think it could be better to have a class which is available in CR (instead of copy-pase the whole mixin)
- removes `overflow-x` on RichText - if link was a last child within RichText it was showing the vertical scroll (due to adding padding-bottom to anchor)

### Test

Link to a QA branch: https://qa-8554-yellow-links.wellcome-corporate-fe.org.uk/

Sorry for a big PR, I hope looking at this page with components make it easier to see changes: https://qa-8554-yellow-links.wellcome-corporate-fe.org.uk/what-we-do/our-work/listing-options.

We can compare the changes to this link: https://develop.wellcome-corporate-fe.org.uk/what-we-do/our-work/listing-options

### DIFF
--- a/src/assets/styles/10-settings/_colours.scss
+++ b/src/assets/styles/10-settings/_colours.scss
@@ -50,9 +50,9 @@
 
   --link-colour: var(--colour-blue-60);
   --link-colour-hover: var(--colour-blue-70);
-  --link-colour-focus-outline: var(--colour-blue-40);
+  --link-colour-focus-outline: var(--colour-blue-50);
   --link-colour-active: var(--colour-blue-50);
-  --link-colour-disabled: var(--colour-grey-50);
+  --link-colour-disabled: var(--colour-grey-40);
 
   --body-text-colour: var(--colour-grey-80);
   --meta-text-colour: var(--colour-grey-60);
--- a/src/assets/styles/10-settings/_colours.scss
+++ b/src/assets/styles/10-settings/_colours.scss
@@ -48,10 +48,10 @@
   --colour-red-90: #40120d;
   --colour-red-60: #e10f2d;
 
-  --link-colour: var(--colour-blue-60);
-  --link-colour-hover: var(--colour-blue-70);
+  --link-colour: var(--colour-grey-80);
+  --link-colour-hover: var(--colour-grey-80);
   --link-colour-focus-outline: var(--colour-blue-50);
-  --link-colour-active: var(--colour-blue-50);
+  --link-colour-active: var(--colour-grey-80);
   --link-colour-disabled: var(--colour-grey-40);
 
   --body-text-colour: var(--colour-grey-80);
--- a/src/assets/styles/20-tools/_mixins/_typography.scss
+++ b/src/assets/styles/20-tools/_mixins/_typography.scss
@@ -99,6 +99,7 @@
   $underline-colour: var(--colour-grey-80),
   $selector: null
 ) {
+  @extend %anchor-base;
 
   text-decoration: none;
 
--- a/src/assets/styles/20-tools/_mixins/_typography.scss
+++ b/src/assets/styles/20-tools/_mixins/_typography.scss
@@ -54,16 +54,16 @@
  * i.e. hover-underline adds styles to a title after the image was hovered
  */
 @mixin default-underline($height: 3px, $underline-colour: var(--colour-grey-80)) {
-  --transition-duration: 0.35s;
+  --transition-duration-link: 0.3s;
   --transition-timing-function: ease-in-out;
 
   background-image: linear-gradient($underline-colour, $underline-colour), linear-gradient(transparent, transparent);
   background-position: left calc(100% - #{$height});
   background-repeat: no-repeat;
   background-size: 100% 1px, 0 $height;
-  padding-bottom: 5px;
+  padding-bottom: 6px;
 
-  transition: var(--transition-duration) var(--transition-timing-function);
+  transition: var(--transition-duration-link) var(--transition-timing-function);
   transition-property: background-size;
 }
 
@@ -99,8 +99,6 @@
   $underline-colour: var(--colour-grey-80),
   $selector: null
 ) {
-  --transition-duration: 0.35s;
-  --transition-timing-function: ease-in-out;
 
   text-decoration: none;
 
--- a/src/assets/styles/20-tools/_mixins/_typography.scss
+++ b/src/assets/styles/20-tools/_mixins/_typography.scss
@@ -49,6 +49,93 @@
 }
 
 /**
+ * Generates a style where a horizontal line displays
+ * beneath an element and changes to another colour on hover with a transition.
+ *
+ * @param {Int} $height - height of the "underline"
+ * @param {string} $underline-colour - colour of the "underline"
+ * @param {string} $animated-colour - colour of the "underline" on hover
+ */
+@mixin animated-double-underline($height: 3px, $underline-colour: var(--colour-grey-80), $animated-colour: var(--colour-amber-30)) {
+  --transition-duration: 0.3s;
+  --transition-timing-function-in: cubic-bezier(0.23, 1, 0.32, 1);
+  --transition-timing-function-out: cubic-bezier(0.55, 0.085, 0.68, 0.53);
+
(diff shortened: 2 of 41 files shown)


src/components/Pagination/_pagination.scss
```diff
@@ -38,11 +38,20 @@
 .cc-pagination__prev-link,
 .cc-pagination__link {
   align-items: center;
+  color: var(--colour-blue-60);
   cursor: pointer;
```
skibinska, line 42: We are keeping the old blue styles for pagination (this will be updated later).

src/assets/styles/50-utilities/_links.scss
```diff
@@ -0,0 +1,11 @@
+// ----------------------------------
+// UI Settings
+// Links
+// ----------------------------------
+// Design System 1.0 Alpha
+// 2019-11-11
+// ----------------------------------
+
+.u-animated-link {
```
skibinska, line 9: I added this class to reuse it in CR (instead of copy-pasting the whole mixin)

src/components/Button/_button.scss
```diff
@@ -26,97 +26,93 @@
 .btn--secondary,
 .btn--ghost {
```
creido-welly: To avoid confusion between different platforms this class should be renamed as `.btn--tertiary`. Likewise all associated classes, variables and props should use `tertiary` not `ghost`.

src/components/Button/_button.scss
```diff
@@ -26,97 +26,93 @@
...
   }
+}
 
-  &[disabled] {
-    border-color: var(--link-colour-disabled);
+.btn--ghost {
+  background-color: transparent;
+
+  &:hover {
+    background-color: var(--colour-amber-20);
```
creido-welly: Sketch has the hover background for this as `--colour-amber-05`

src/assets/styles/20-tools/_mixins/_typography.scss
```diff
@@ -48,6 +48,86 @@
...
+ * }
+ * or
+ * a {
+ *  @include animated-link;
+ * }
+ *
+ * @param {Int} $height - height of the "underline" on hover
+ * @param {string} $hover-colour - colour of the "underline" on hover
+ * @param {string} $underline-colour - colour of the "underline" (default state)
+ * @param {string} $selector - selector to apply anchor styles
```
creido-welly: `$child-selector` seems like it would be a better name for this

src/components/InpageNav/_inpage-nav.scss
```diff
@@ -53,27 +53,25 @@
 }
 
 .cc-inpage-nav__link {
-  border-left: 2px solid transparent;
-  color: var(--colour-grey-60);
+  @include animated-link($selector: 'span');
```
creido-welly: I assume you've done this because mixing the animated link and inpage nav link styles on the same element has side effects?

src/assets/styles/50-utilities/_links.scss
creido-welly: I think that makes sense given the use case. Applying `@include animated-link` explicitly on every anchor (e.g.`.cc-card__link`) or component which has anchor children (e.g. `.cc-rich-text a`) probably gives us better overall style control and of course there is less call for overriding it. The drawback with that is style duplication.

If animated underline was applied globally instead (on `a { ... }` - in CR basically) it would cascade down to every instance. Easier but potentially very annoying if overrides are ever needed.

You're closest to this, what do you think?

src/assets/styles/10-settings/_colours.scss
```diff
@@ -48,11 +48,11 @@
   --colour-red-90: #40120d;
   --colour-red-60: #e10f2d;
 
-  --link-colour: var(--colour-blue-60);
-  --link-colour-hover: var(--colour-blue-70);
-  --link-colour-focus-outline: var(--colour-blue-40);
-  --link-colour-active: var(--colour-blue-50);
-  --link-colour-disabled: var(--colour-grey-50);
+  --link-colour: var(--colour-grey-80);
```
creido-welly, line 51: Ultimately we need to refactor and expand this set of variables to include background colours (which these were originally) but I think that's something to be tackled in the design system code.

src/components/Button/_button.scss
```diff
@@ -26,97 +26,93 @@
...
+  }
 
   &:active {
-    background-color: var(--colour-blue-10);
+    background-color: var(--colour-amber-10);
   }
 
-  &[disabled] {
-    background-color: var(--colour-grey-20);
+  .btn__text {
```
creido-welly: This child `.btn__text` and the following state selectors should be flattened as it's becoming harder to read here

i.e. 

```
.btn--tertiary .btn__text {
  ...
}

.btn--tertiary:hover .btn__text,
.btn--tertiary:active .btn__text,
.btn--tertiary[disabled] .btn__text {
  ...
}
```

creido-welly (changes requested):
There's a lot to take in here but I think I've been through most of it. My main query is regarding the approach to applying of the animated-link mixin - [global or component level](https://github.com/wellcometrust/corporate-components/pull/494/files#r644013236).

src/assets/styles/50-utilities/_links.scss
skibinska: @creido-welly 
Yes, if we apply underline globally in CR we will have to override a lot of links styles in the components. I read someone that if we have to undo a lot of styles it means there were added too soon (I was looking for that article, but can't find it now).

I left globally the `%anchor-base` styles on anchor in CR, in case I missed any link. (we are making sure that links have distinguish style even if missed out, and it should be easier to apply underline, once we spot them).

src/components/InpageNav/_inpage-nav.scss
skibinska: Without span the inpageNav links don't look correct.

Screenshot 2021-06-07 at 12 03 09

Should I try to add these styles without updating markup?

src/assets/styles/10-settings/_colours.scss
skibinska: I will create a ticket to add variables for the button background-colours.
Ticket created: https://github.com/wellcometrust/corporate/issues/8631

src/components/InpageNav/_inpage-nav.scss
creido-welly: I would be inclined to use whichever way is less confusing.

src/assets/styles/50-utilities/_links.scss
creido-welly: Ok. We can review it in future and decide whether to keep as it is or change.

As an aside, SASS placeholders have been completely removed in design system. Where necessary they're being replaced with mixins.
https://github.com/wellcometrust/design-system/tree/develop/src/styles/20-tools

src/assets/styles/50-utilities/__manifest.scss
```diff
@@ -7,5 +7,6 @@
 
 @import 'a11y';
 @import 'external-link-indicator';
+@import 'links';
```
creido-welly, line 11: Minor: alphabetical order

src/components/RichText/_rich-text.scss
```diff
@@ -109,6 +112,11 @@
     @extend %body-regular;
   }
 
+  a {
```
creido-welly: Duplication of lines 49-52

src/components/InpageNav/_inpage-nav.scss
skibinska: @creido-welly if that's okay with you, I would like to keep it with span for now.

creido-welly (approved):

creido-welly (approved):

PR #497, opened 2021-05-28
8617 bug: link list incorrect markup

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8617

### Context

List link markup was incorrect - it was displaying `<div>` instead of `<ul>`.

I have made a change based on assumption that `link_list` variant is wrapping the `<li>` elements.

I created a [page, where I listed the list variants](https://develop.wellcome-corporate-fe.org.uk/what-we-do/our-work/listing-options).

### This PR

- adds conditional to check if the `list_link` option was selected then use `<ul>` as a wrapper

### DIFF
--- a/src/components/InfoBox/InfoBox.stories.tsx
+++ b/src/components/InfoBox/InfoBox.stories.tsx
@@ -40,10 +40,8 @@ const InfoBoxExample = () => {
           priority for G20 countries, philanthropy and the private sector.
         </p>`}
       </Text>
-      <Listing>
-        <Contact name="John Smith" />
-        <Contact name="Bertie Basset" />
-      </Listing>
+      <Contact name="John Smith" />
+      <Contact name="Bertie Basset" />
       <Listing>
         <ListingLink href="#1" title="Link 1 text" />
         <ListingLink href="#2" title="Link 2 text" />
--- a/src/components/Listing/Listing.tsx
+++ b/src/components/Listing/Listing.tsx
@@ -37,7 +37,7 @@ export const Listing = ({
   className,
   variant = 'link_list'
 }: ListingProps) => {
-  const Element = as;
+  const Element = variant === 'link_list' ? 'ul' : as;
   const classNames = cx(variantMapping[variant], {
     [`${className}`]: className
   });


src/components/InfoBox/InfoBox.stories.tsx
```diff
@@ -40,10 +40,8 @@ const InfoBoxExample = () => {
           priority for G20 countries, philanthropy and the private sector.
         </p>`}
       </Text>
-      <Listing>
-        <Contact name="John Smith" />
-        <Contact name="Bertie Basset" />
-      </Listing>
+      <Contact name="John Smith" />
```
skibinska, line 43: I don't think it is possible to add a Contact component within Listing.

(I based it on [the page I created](https://develop.wellcome-corporate-fe.org.uk/what-we-do/our-work/listing-options) and on the teams page where there is a lot of contact us within InfoBox and non of it is wrapped within `<Listing>` component, e.g. [International cultural programmes team](https://wellcome.org/who-we-are/teams/international-cultural-programmes-team)

src/components/Listing/Listing.tsx
```diff
@@ -37,7 +37,7 @@ export const Listing = ({
   className,
   variant = 'link_list'
```
skibinska, line 38: I was wondering if we need this to be default, but I guess when we create a custom page and want to add a list of links, it could be easier to just add `<Listing>` without any props for the correct markup to be created.

src/components/Listing/Listing.tsx
```diff
@@ -37,7 +37,7 @@ export const Listing = ({
...
 }: ListingProps) => {
-  const Element = as;
+  const Element = variant === 'link_list' ? 'ul' : as;
```
skibinska, line 40: I am making an assumption that we want to create an unordered list. I don't think DCPs have an option to choose the wrapping element and I didn't see `<ol>` as a wrapper for a link list anywhere in the codebase.

src/components/InfoBox/InfoBox.stories.tsx
creido-welly: It may have been the case once upon a time but looking at the CMS I don't think it is possible.

src/components/Listing/Listing.tsx
creido-welly: I _think_ it's a safe call but it would be good to test by creating/testing each variant, just in case.

creido-welly (approved):
Approved pending testing against each variant to ensure markup is correct

skibinska:
Thanks @creido-welly! I added these changes to my QA links branch. I added a page where I listed all of the listing options and I can confirm that everything looks ok :) 

Link to qa branch with listed listing options: https://qa-8554-yellow-links.wellcome-corporate-fe.org.uk/what-we-do/our-work/listing-options.

We can compare it to the develop version, where markup is incorrect: https://develop.wellcome-corporate-fe.org.uk/what-we-do/our-work/listing-options

creido-welly (approved):
All good to merge

PR #510, opened 2021-07-23
8807 task: adds captionDisplay option to Image component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8807

### Context

Currently, when editors add image, caption displays on the right hand side of the image. For some content it does not work. So we need different display option for captions.

In the backend there are 4 different options now:

- **None** -  shows caption on the right 
-  **Right column (default)** - shows caption on the right
- **Below image** - caption below image
- **Hide caption** - hides caption but show meta information (we need this option to prevent creating multiply images with different info)

More info here: https://github.com/wellcometrust/corporate/issues/8768

### This PR

- adds `captionDisplay` option to Media and Image components

### DIFF
--- a/src/components/Image/Image.tsx
+++ b/src/components/Image/Image.tsx
@@ -5,11 +5,11 @@ import ImageElement from './ImageElement';
 
 type ImageProps = {
   alt: string;
-  caption?: string;
+  captionText?: string;
   captionVariant?: 'below' | 'right';
   credit?: string;
   className?: string;
-  hideCaption?: boolean;
+  shouldHideCaption?: boolean;
   licence?: string;
   sizes?: string;
   src: string;
@@ -21,23 +21,23 @@ const imageSizesDefault =
 
 export const Image = ({
   alt = '',
-  caption,
+  captionText,
   captionVariant = 'right',
   credit,
   className,
-  hideCaption,
   licence,
+  shouldHideCaption,
   sizes = imageSizesDefault,
   src,
   srcSet
 }: ImageProps) => (
   <Media
-    caption={caption}
+    caption={captionText}
     captionVariant={captionVariant}
     className={className}
     credit={credit}
-    hideCaption={hideCaption}
     licence={licence}
+    shouldHideCaption={shouldHideCaption}
   >
     <ImageElement alt={alt} sizes={srcSet && sizes} src={src} srcSet={srcSet} />
   </Media>
--- a/src/components/Image/Image.tsx
+++ b/src/components/Image/Image.tsx
@@ -6,6 +6,7 @@ import ImageElement from './ImageElement';
 type ImageProps = {
   alt: string;
   caption?: string;
+  captionDisplay?: string;
   credit?: string;
   className?: string;
   licence?: string;
@@ -20,6 +21,7 @@ const imageSizesDefault =
 export const Image = ({
   alt = '',
   caption,
+  captionDisplay,
   credit,
   className,
   licence,
@@ -29,6 +31,7 @@ export const Image = ({
 }: ImageProps) => (
   <Media
     caption={caption}
+    captionDisplay={captionDisplay}
     className={className}
     credit={credit}
     licence={licence}
--- a/src/components/Image/Image.tsx
+++ b/src/components/Image/Image.tsx
@@ -6,9 +6,10 @@ import ImageElement from './ImageElement';
 type ImageProps = {
   alt: string;
   caption?: string;
-  captionDisplay?: string;
+  captionVariant?: 'below' | 'right';
   credit?: string;
   className?: string;
+  hideCaption?: boolean;
   licence?: string;
   sizes?: string;
   src: string;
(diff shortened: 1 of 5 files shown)


src/components/Image/Image.tsx
```diff
@@ -6,6 +6,7 @@ import ImageElement from './ImageElement';
 type ImageProps = {
   alt: string;
   caption?: string;
+  captionDisplay?: string;
```
chriddell: You could declare explicit values here, because they are known (`below_image || hide || ...`). Allowing any string value could cause unexpected behaviour for somone who is trying to use the component (e.g. they provide captionDisplay = "belowImage"). 

If you declare the potential values explicitly, a type error will be thrown to let the person know the prop only accepts specific values.

chriddell (changes requested):
No issues with the code (other than the one comment I left on type values).

One thought: is it appropriate to use values from Drupal in this repo? The intention of this PR is to allow the caption for an image to display on the right of, or below, an Image (or not appear at all!).

To do this, you're checking for a value of `below_image` which is a value set by Drupal. If I understand the purpose of corporate-components correctly, it shouldn't be tying implementation details to Drupal values. It means that if something changes in Drupal (e.g. the value of "hide" becomes "should_hide") then we need to change the code in corporate-components, too - annoying!

It's a case of "the tail wagging the dog", I think, where Drupal is determining what corporate-components should do.

My advice would be to think about what you want the component here to do, and implement that in the "best" way possible, and allow corporate-react to do the translation between Drupal and corporate-components.

In relation to changes to this PR, I think you could:

- add a prop `hideCaption` (boolean)
- allow two values for `captionDisplay` ("right" | "below") (you could even abstract this to "standard" | "alternative" - or something like that - if we wanted to be REALLY agnostic)
- set default value of `captionDisplay` to "below"

I think that would acheive the same thing, but be less tied to Drupal. Then corporate-react could do the translation of `below_image` -> `below` etc.

Does that make sense?? Let me know if not... this also assumes that I understand the direction of corporate-components - that it _shouldn't_ be tied to the Drupal site, and will eventually be used by other consumers (e.g. the funding platform).

skibinska:
Thank you @chriddell, that's a great point. I will update this (and change the prop name `captionDisplay` to `captionVariant`).

src/components/Media/Media.tsx
```diff
@@ -5,17 +5,21 @@ import MediaCaption from './MediaCaption';
...
   children: React.ReactElement;
   className?: string;
   credit?: string;
+  hideCaption?: boolean;
   licence?: string;
 };
 
 const Media = ({
   caption,
+  captionVariant = 'right',
```
skibinska, line 18: I decided to use `right` as a default as this is our default for images and videos in CR.

skibinska:
@chriddell this is ready to review again :)

src/components/Image/Image.tsx
```diff
@@ -6,8 +6,10 @@ import ImageElement from './ImageElement';
 type ImageProps = {
   alt: string;
   caption?: string;
+  captionVariant?: 'below' | 'right';
```
creido-welly, line 9: `captionVariant` works but doesn't immediately communicate what the prop is for whereas `captionPosition` definitely would.

src/components/Image/Image.tsx
```diff
@@ -6,8 +6,10 @@ import ImageElement from './ImageElement';
 type ImageProps = {
   alt: string;
   caption?: string;
```
chriddell: Now that we have `captionVariant`, maybe `caption` should become `captionText` to avoid confusion?

src/components/Image/Image.tsx
```diff
@@ -6,8 +6,10 @@ import ImageElement from './ImageElement';
...
+  captionVariant?: 'below' | 'right';
   credit?: string;
   className?: string;
+  hideCaption?: boolean;
```
chriddell: To align with other boolean prop names, I think we should use the same convention. So this could be `shouldHideCaption` instead.

I just looked this up, and apparently these words are called "modal auxilliary verbs" in English grammar 😎 and there are 9 of them: shall, should, can, could, will, would, may, must, might. We should always use these when naming boolean props, I think, because it makes it very explicit to the reader that this is a boolean, and once we're used to it I don't think it will require extra effort.

https://www.google.com/search?q=modal+auxiliary+verb

src/components/Media/Media.tsx
chriddell: Good consistency

chriddell (approved):
Left a couple of comments amount renaming variables, but other than that I think this is all good. I'll approve so that you can merge when you have addressed the other comments.

src/components/Image/Image.tsx
chriddell: Could argue that `captionVariant` allows for differences in style, that may be implemented based on the position? (Point being that maybe `captionVariant` gives more scope, which may be beneficial)

creido-welly: Ok

PR #513, opened 2021-08-03
8813 task: adds person component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8813

### Context 

User story: **As a user, I want to see more faces of people who work at Wellcome, so that I can trust Wellcome.**

Link to design: https://www.sketch.com/s/8ee1a7e8-969a-4e6c-875f-2e5333a58605

### This PR

- adds `Person` component
- adds styles for `author` and `full-width` options
- adds Person story

### Test

- fetch this branch and run `npm run storybook`

**Full width option:**

Screenshot 2021-08-03 at 16 08 33

**Author option:**

Screenshot 2021-08-03 at 16 10 08

### DIFF
--- a/src/components/Person/Person.tsx
+++ b/src/components/Person/Person.tsx
@@ -50,15 +50,18 @@ export const Person = ({
     )}
     <div className="cc-person__wrapper">
       <div className="cc-person__header">
-        <h3 className="cc-person__title" itemProp="name">
+        <h3 className="cc-person__heading" itemProp="name">
           {name}
         </h3>
-        <p className="cc-person__title cc-person__subtitle" itemProp="jobTitle">
+        <p
+          className="cc-person__heading cc-person__subheading"
+          itemProp="jobTitle"
+        >
           {jobTitle}
         </p>
         {organisation && (
           <p
-            className="cc-person__title cc-person__subtitle"
+            className="cc-person__heading cc-person__subheading"
             itemProp="memberOf"
           >
             {organisation}
--- a/src/components/Person/Person.tsx
+++ b/src/components/Person/Person.tsx
@@ -9,7 +9,7 @@ export type PersonProps = {
   imageSizes?: string;
   imageSrc?: string;
   imageSrcSet?: string;
-  jobTitle: string;
+  jobTitle?: string;
   layoutVariant: 'author' | 'full-width' | 'narrow' | 'team';
   links?: {
     title: string;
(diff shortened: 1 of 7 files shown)


src/components/Person/_person.scss
```diff
@@ -0,0 +1,116 @@
...
+  @include mq(sm) {
+    display: flex;
+  }
+
+  a {
+    @include animated-link;
+  }
+}
+
+.cc-person__image {
```
creido-welly, line 26: There's an issue with the image width being squashed > 768px. Applying `flex-shrink: 0;` will force the image width to remain constant. Also by default the figure element this selector is applied to looks like it could use some margin resetting. The excess left hand margin is very noticeable on mobile.

Screenshot 2021-08-03 at 17 29 21

src/components/Person/Person.tsx
```diff
@@ -0,0 +1,96 @@
...
+          alt={`A photograph of the person, ${name}.`}
+        />
+      </figure>
+    )}
+    <div className="cc-person__wrapper">
+      <div className="cc-person__header">
+        <h3 className="cc-person__heading" itemProp="name">
+          {name}
+        </h3>
+        <p
```
creido-welly: I can see from the type definition that jobTitle is expected but is there any likelihood of this being an optional value? Might it need logic?

skibinska: Sure, I can add a logic to this, but at this point it is not possible to create a new Person without a job title (it is required field in the CMS).

Link to the CMS: https://wt-corporated8-develop.codeenigma.net/node/add/person

src/components/Person/_person.scss
```diff
@@ -0,0 +1,116 @@
...
+  @extend %heading-large;
+
+  margin-bottom: 0;
+}
+
+.cc-person__subheading {
+  font-weight: normal;
+}
+
+.cc-person__description {
```
creido-welly, line 69: I noticed [global line-height for Storybook](https://github.com/wellcometrust/corporate-components/blob/9871d3a46547a50579aac1ee8e074f072a2d3895/config/storybook/styles/_storybook-global.scss#L11) is using an old CC variable name `--body-line-height-md`. It should be updated to `--body-line-height`.

Line height should probably also be set here too since we're doing this for other components.

skibinska: updated :)

creido-welly (approved):

PR #514, opened 2021-08-05
8843 task: adds PeopleList component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8843

### Context

Adds compact variant to People component and uses it to display a list of people (it can be use to show team members).

Link to design: https://www.sketch.com/s/8ee1a7e8-969a-4e6c-875f-2e5333a58605

### This PR

- adds `PeopleList` component
- updates `layoutVariant` in People component

### Test

- fetch this branch and run `npm run storybook`

Example: **Grid variant**
Screenshot 2021-08-05 at 12 32 56


src/components/Person/Person.tsx
```diff
@@ -10,7 +10,7 @@ export type PersonProps = {
   imageSrc?: string;
   imageSrcSet?: string;
   jobTitle?: string;
-  layoutVariant: 'author' | 'full-width' | 'narrow' | 'team';
+  layoutVariant: 'author' | 'compact' | 'full-width' | 'narrow';
```
chriddell: The values "compact" and "narrow" seem confusing (very similar meanings) - maybe a better combination could be used here to make the separation more clear?

chriddell: Also, seems to me that the values "narrow" and "full-width" are never used (I can't see any CSS selectors or conditional statements related to these values?).

If that's the case, could we remove these values to make the code slightly more simple?

src/components/PeopleList/PeopleList.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import Person, { PersonProps } from 'Person/Person';
+
+type PersonPropsExtend = PersonProps & {
+  id: string;
+};
+
+export type PeopleListProps = {
+  people: PersonPropsExtend[];
+  variant?: 'grid' | 'list';
```
chriddell: Would using `layoutVariant` (instead of `variant`) be appropriate? It would match the name of the prop in the Person component, which I think is serving the same purpose.

chriddell (commented):
It seems like the PeopleList component is a fairly straightforward `<ul>` with some layout (flex) CSS applied. To me it seems like a lot of overhead to maintain a component this specific in the corporate-components repo.

It would seem more appropriate for these styles to be applied in corporate-react, I think. The decision of how to lay out the Person components is then decided there, and it means we could lay out the Person components differently in different apps which use corporate-components.

My preference would be to do something like this in corporate-react:

```css
.u-flex {
  display: flex;
}

.u-flex-wrap {
  flex-wrap: wrap;
}

/* You could add breakpoint variants with something like... */

@each $breakpoint in $mq-breakpoints {
  .u-flex@#{$breakpoint} {
    @include mq($breakpoint) {
      display: flex;
    }
  }
}

/* I think there may already be a width: 50%; utility class but if not, add one */
```

and then wherever we need to render a list of people

```js
<ul className="o-cleanlist u-flex u-flex-wrap">
  {people.map(props => <Person className="u-mb-XX" {...props} />}
</ul>
```

Unless! Maybe you have already considered this approach, and you decided to add this component in corporate-components for other reasons?

src/components/Person/Person.tsx
skibinska: The only option I am not using is `full-width`, which is a default. I am getting this option from the CMS.  I can actually add some checks in corporate-react to pass an empty string.

The options that I am getting from the CMS are: 

- full-width (default styles) - width: 9 columns
- narrow - very similar to default, but has different layout on the small desktop (width: 6 columns)
- author - yellow background, smaller font
- team - which is used to create a people list
- we are missing one more option in the CMS to add people grid layout

[All the widths are handled in the CR](https://github.com/wellcometrust/corporate-react/blob/develop/src/utils/set-layout-variant.ts#L31).

 `team` option (and `grid` option):

I decided to add a `compact` option (which can be used to add a person to a quote and will replace an Author) as `team` and `grid`(or whatever this option is named) are not a variants for the person component, they are variants for displaying multiple person components.

I am happy to rename any of these options.

skibinska:
@chriddell I am happy to move this to corporate-react.

src/components/Person/Person.tsx
chriddell: Okay cool, thanks for the explainer!

- if full-width is the default, let's make that clear (set default prop value to "full-width")
- I do think having "narrow" and "compact" is confusing... if "compact" is related to being inside a Quote component, could we name it in a way that makes this clear? e.g. "quote" 

I think that should make things a little clearer?

skibinska: 1. I actually got rid of the full-width/default, so I am passing an empty string,  but maybe I should pass a `null`.
2. I can change "compact" to "quote", but just wanted to flag that it is going to be used not only inside Quote, but as well for a list of people (e.g. team list) and **maybe**  instead of an Author component at the top of the article.

[Person component - sketch file](https://www.sketch.com/s/8ee1a7e8-969a-4e6c-875f-2e5333a58605)

chriddell: 1. Null would be better, otherwise we could run in to issues if an empty string is passed through to the component as it may be not be omitted as falsy... (I think ReactDOM is smart enough to prevent empty strings being passed as values for props, but I can't find any documentation to support my thinking!). Null would be safer, anyway.

2. Ah, okay, don't change to "quote" then. I think we cannot have "narrow" and "compact", though, as they are too similar. Unless these words were decided by the team and they are what designers use? If that's the case, then that's okay, but I think really they should be less similar words.

skibinska: No, these words were not decided by team, and I am happy to change them.

chriddell: Cool... should we close this PR, and continue the chat on https://github.com/wellcometrust/corporate-react/pull/1050? I'm getting a bit confused with the different PRs! And I think we can close this one?

PR #518, opened 2021-08-18
8679 bug: updates accordion font-weight

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8679

### Context

As a part of updating the font on Windows machine Jason requested to change the accordions to `font-weight: 500`

### This PR

- updates accordion's `font-weight`

### DIFF
--- a/config/storybook/styles/_storybook-global.scss
+++ b/config/storybook/styles/_storybook-global.scss
@@ -65,3 +65,7 @@ h6 {
     transform: translate3d(0, 64px, 0);
   }
 }
+
+.cc-accordion__button {
+  font-family: var(--font-primary);
+}
--- a/config/storybook/styles/_storybook-global.scss
+++ b/config/storybook/styles/_storybook-global.scss
@@ -65,7 +65,3 @@ h6 {
     transform: translate3d(0, 64px, 0);
   }
 }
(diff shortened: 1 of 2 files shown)


config/storybook/styles/_storybook-global.scss
```diff
@@ -65,3 +65,7 @@ h6 {
     transform: translate3d(0, 64px, 0);
   }
 }
+
+.cc-accordion__button {
```
skibinska: without this style the accordion button in storybook has font Arial. 
@creido-welly, is it okay for this to be here, or should I move it to `_accordion.scss`

creido-welly: There's obviously a global style applied to the site which isn't present in SB. Try adding it to the accordion itself and see if that also fixes it in storybook. If not you may need to specify it for this selector in `_accordion.scss`. You shouldn't have to add a fix for this selector in SB.

skibinska: I added `font-family: inherit` to the accordion-button - thanks to this the correct font is used in storybook (Helvetica Neue) and for Windows users (Inter).

creido-welly (approved):

PR #521, opened 2021-08-31
8912 task: removes SiteAlert and WellcomeCollectionBanner

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8912

### This PR

- removes SiteAlert
- removes unused WellcomeCollectionBanner


src/assets/styles/30-components/__manifest.scss
```diff
@@ -72,4 +72,3 @@
 @import './src/components/TextInput/text-input';
 @import './src/components/Timeline/timeline';
 @import './src/components/Video/video';
-@import './src/components/WellcomeCollectionBanner/wellcome-collection-banner';
```
creido-welly, line 75: I assume WellcomeCollectionBanner is now completely unused, hence you removing it?

src/components/Link/Link.tsx
```diff
@@ -24,7 +24,7 @@ const isFileLink = (str: string) =>
  * @param { string } url
  * @returns { boolean }
  */
-const isExternalLink = (url: string) => {
```
creido-welly: Since this is purely a JavaScript utility and it's being used in multiple places now it would be better off being moved to the utils folder

creido-welly (commented):
Everything in this PR looks fine on the whole with a couple of minor things to address.

Ideally the SiteAlert component needs moving to corporate-react at some point. It is pretty site specific and I can't see it moving to design system any time soon if at all. You can either remove it in this PR and create a separate PR for adding it into corporate-react or add a `deprecated` notice similar to [NewsletterSignup](https://github.com/wellcometrust/corporate-components/blob/1748f08e2cbff6b7958884a83c6d23fa498ee1e6/src/components/NewsletterSignup/NewsletterSignup.tsx#L1) and address that job as separate PRs.

src/components/SiteAlert/SiteAlert.tsx
```diff
@@ -23,19 +25,33 @@ export const SiteAlert = ({
...
+  const [isExternal, setExternal] = useState(false);
+
+  // Ensures 'window' is present
+  useEffect(() => {
+    if (url) {
+      setExternal(isExternalLink(url));
+    }
+  }, []);
+
   const tabIndex = isActive ? 0 : -1;
```
creido-welly: I've not seen it mentioned in the audit but it's worth noting that the SiteAlert close button and link are still accessible from the VoiceOver rotor (see https://webaim.org/articles/voiceover/#navigation for further info on accessing that if you're not familiar with it). Should be able to fix that by adding `aria-hidden` to them in a similar way to how `tabIndex` is applied e.g.

```
const ariaHidden = isActive ? false : true;
```

src/assets/styles/30-components/__manifest.scss
skibinska: I didn't find it anywhere in the codebase. We are using SiteAlert for Wellcome Collection banner in CR.

creido-welly (approved):

PR #522, opened 2021-08-31
8900 task(style): minor person component updates

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8900

### This PR

- changes person image `background-color`
- updates person image size on large desktop

### DIFF
--- a/src/components/PageHeaderCompact/PageHeaderCompact.tsx
+++ b/src/components/PageHeaderCompact/PageHeaderCompact.tsx
@@ -84,7 +84,6 @@ export const PageHeaderCompact = ({
         <div className="cc-page-header-compact__authors">
           {authors.map(author => (
             <Author
-              imageSizes="(min-width: 48em) 6.25rem, 3.125rem"
               imageSrc={author.imageSrc}
               imageSrcSet={author.imageSrcSet}
               jobTitle={author.jobTitle}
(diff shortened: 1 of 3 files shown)


src/components/PageHeaderCompact/PageHeaderCompact.tsx
```diff
@@ -84,7 +84,6 @@ export const PageHeaderCompact = ({
         <div className="cc-page-header-compact__authors">
           {authors.map(author => (
             <Author
-              imageSizes="(min-width: 48em) 6.25rem, 3.125rem"
```
creido-welly, line 87: Out of interest, why are you removing this?

It would make a lot sense to update it to the new author image sizes i.e. 300, 200, 160, 100 and finally 80 and adding in the breakpoints that each kicks in at.

e.g. 
```js
imageSizes="
  (min-width: 64em) 18.75rem, // 300px at 1024 wide
  (min-width: 48em) Xrem, // Xpx at 768 wide etc
  5rem" // 80px
```

skibinska: This is for the author(s) at the top of an article. An image here has the same size across all breakpoints (80px)

Screenshot 2021-09-01 at 10 19 14

creido-welly (approved):

PR #529, opened 2021-09-15
8957 task: removes newsletter pop-up error

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8957

### Context

When tabbing past the ‘Your email address’ input field error handling is triggered and an error occurs; however, this is not conveyed to screen reader users at this point. Remove the pop-up error altogether and only trigger error handling once the users makes an incorrect submission by inputting the incorrect email address format.

### This PR

- shows error message only after the incorrect email address format is used
- adds `aria-describedby` and id to convey error message to screen reader user

### DIFF
--- a/src/components/NewsletterForm/NewsletterFormEmail.tsx
+++ b/src/components/NewsletterForm/NewsletterFormEmail.tsx
@@ -32,7 +32,7 @@ export const NewsletterFormEmail = ({
         Your email address
       </label>
       <input
-        aria-describedby={hasError && `${id}-error`}
+        aria-describedby={hasError ? `${id}-error` : null}
         autoComplete="email"
         className={classNames}
         id={id}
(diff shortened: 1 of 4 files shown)


src/components/NewsletterForm/NewsletterFormEmail.tsx
```diff
@@ -32,6 +32,7 @@ export const NewsletterFormEmail = ({
         Your email address
       </label>
       <input
+        aria-describedby={hasError && `${id}-error`}
```
creido-welly: This attribute expects an id or a space separated list of ids. If hasError is false then `false` would be output as the attribute value so I imagine the following would be better

```
aria-describedby={hasError ? `${id}-error` : null}
```

creido-welly (approved):

PR #534, opened 2021-09-21
8303 task: Upgrade CC to node 15 - HTML sanitising and parsing refactor

creido-welly:
**Please note, the massive number of line changes (48,001) is as a consequence of the package-lock being updated from lock version 1 to lock version 2**

## Context
In order to integrate design system into Wellcome.org, corporate-components must first be upgraded to the same node version (node 15.0.1, NPM 7.0.3). corporate-react is also on this version. Unfortunately react-html-parser (or one of its dependencies) is incompatible with this version and so this PR features updates to replace it.

## Description
Instead of a straight parser I have opted for a reliable sanitiser library [XSS](https://www.npmjs.com/package/xss) which has really wide adoption and use React's built in dangerouslySetInnerHTML prop to handle the parsing.
 
The only downside to this is where we have a component with a prop such as title into which we want to inject HTML we can't simply do this
```jsx
<MyComponent title={parseHtml(titleText)} />
```

... without handling the escaped HTML within the component itself.
```jsx
const MyComponent =  ({ title }) => (
  <div dangerouslySetInnerHTML={{ __html: titleText }} />
)
```

I would argue we should be doing this anyway if we intend a prop to be able to handle HTML, in which case we should do the sanitising in the component too.
```jsx
const MyComponent =  ({ title }) => (
  // parseHtml no longer appropriate as a name sanitiseHTML
  <div dangerouslySetInnerHTML={{ __html: sanitiseHtml(titleText) }} />
)
```

`sanitiseHTML` util is exported from corporate-components for use in corporate-react.


src/utils/parse-html.ts
```diff
@@ -4,22 +4,37 @@
...
+
+// other libraries which have been tried and found wanting
+
+// import ReactHtmlParser, { Options } from 'react-html-parser';
+// import parser from 'html-react-parser';
+// import sanitizeHtml from 'sanitize-html';
+// import htmlparser2 from 'htmlparser2';
+// import parser from 'parse-entities';
+
+// no server-side, otherwise really good
```
creido-welly: I should have said "no server-side without added hackery"

https://stackoverflow.com/questions/65646007/next-js-dompurify-sanitize-shows-typeerror-dompurify-webpack-imported-module

aliceh75:
This may be a silly question (as I'm not a React person), but is it possible to create custom attributes (with custom functionality) that apply to all components in a React application ?

If that was possible, instead of using `dangerouslySetInnerHtml` we could create a custom attribute that works on all components (say `setRawInnerHtml`) and that one would pass the string through `parseHtml`.

e.g. 
```javascript
const MyComponent =  ({ title }) => (
  <div setRawInnerHtml={{ title }} />
)
```

Where having `setRawInnerHtml` would automatically invoke something like:

```javascript
function setRawInnerHtml(component, text) {
  component.innerHtml = parseHtml(text)
}
```

It would protect against the risk of forgetting to call `parseHtml`. But maybe that approach is not possible with React.

creido-welly:
> This may be a silly question (as I'm not a React person), but is it possible to create custom attributes (with custom functionality) that apply to all components in a React application ?
> 
> If that was possible, instead of using `dangerouslySetInnerHtml` we could create a custom attribute that works on all components (say `setRawInnerHtml`) and that one would pass the string through `parseHtml`.
> ...
> It would protect against the risk of forgetting to call parseHtml. But maybe that approach is not possible with React.

I don't think you could create a custom attribute for an HTML element such as a div but you could create a custom component and set the HTML text as a prop, but it still needs to be parsed somehow so the HTML tags are rendered as markup and not text. The standard React way of doing it is via dangerouslySetInnerHTML and this is what I'm proposing to do with the [RichText component](https://github.com/wellcometrust/corporate-components/pull/534/files#diff-31c13da9e03f05f52afe48b4db5f4d3926764d2a8843097dbaec33172e5e4c49) which we use to ultimately render `paragraph--text` output.

The drawback with this approach is that an extra child element is added each time, which is fine in some cases but I'm not sure if it would be desirable everywhere. It starts to get complicated if we're applying other props to define the element type and class.

An alternative to `dangerous...`, which is what currently happens on live, is to use a third-party utility to parse the HTML which has been sanitised but I surmised it was adding further file size to do something which React already does.

stuartwilkes (approved):
All looks good

creido-welly:
@stuartwilkes thank you

PR #539, opened 2021-10-04
9041 task(style): radio buttons/checkboxes are not discoverable for Dragon users

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/9041

### Context

Accessibility audit:

The customised radio buttons and checkboxes are not discoverable for Dragon users using standard voice command; such as “Click button”.

This issue occurs when the original elements are not hidden correctly with the correct CSS.

[gov.uk radio button example](https://design-system.service.gov.uk/components/radios/default/index.html) -> inspect to see the CSS

### This PR

- hides the checkbox and radio buttons using updated CSS
- created reusable CSS variable for the checkbox and radio buttons size

### DIFF
--- a/src/assets/styles/20-tools/_mixins/_a11y.scss
+++ b/src/assets/styles/20-tools/_mixins/_a11y.scss
@@ -27,3 +27,27 @@
   position: absolute;
   width: 1px;
 }
+
+/**
+ * Hide radio buttons and checkboxes visually (for screen-readers).
+ * Allows Dragon users to access the inputs by using standard voice commands,
+ * such as “Click button”.
+ *
+ * Usage:
+ *
+ * .selector {
+ *   @include visually-hidden-input($top: 0.25rem);
+ * }
+ *
+ */
+
+@mixin visually-hidden-input($top: 0) {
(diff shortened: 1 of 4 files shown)


src/components/Checkbox/_checkbox.scss
```diff
@@ -8,38 +8,37 @@
```
creido-welly, line 8: I think it makes more sense to define `--checkbox-radio-size` in this file

src/assets/styles/20-tools/_mixins/_a11y.scss
```diff
@@ -27,3 +27,27 @@
   position: absolute;
   width: 1px;
 }
+
+/**
+ * Hide radio buttons and checkboxes visually (for screen-readers).
+ * Allows Dragon users to access the inputs by using standard voice commands,
```
creido-welly: Please could you add a reference link (or links if there are several useful resources) for Dragon to this comment block so other developers can easily find out about it

creido-welly (approved):
Looks good, just two minor comments

PR #540, opened 2021-10-04
9031 task: design system colours

creido-welly:
## Context
This PR is one of many to follow in which existing CSS variables are being replaced with the equivalent from the design system.

Also features sass:math usage to remove deprecation warnings.

### DIFF
--- a/src/assets/styles/10-settings/_colours.scss
+++ b/src/assets/styles/10-settings/_colours.scss
@@ -7,15 +7,6 @@
 // ----------------------------------
 
 :root {
-  --colour-black: #000;
-  --colour-white: #fff;
-
-  --colour-amber-40: #e7ae04;
-  --colour-amber-30: #fec200;
-  --colour-amber-20: #ffe699;
-  --colour-amber-10: #fff4d1;
-  --colour-amber-05: #fffbf0;
-
   --colour-blue-90: #002e45;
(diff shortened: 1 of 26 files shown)


src/assets/styles/10-settings/_colours.scss
```diff
@@ -7,15 +7,6 @@
...
-
-  --colour-amber-40: #e7ae04;
-  --colour-amber-30: #fec200;
-  --colour-amber-20: #ffe699;
-  --colour-amber-10: #fff4d1;
-  --colour-amber-05: #fffbf0;
-
   --colour-blue-90: #002e45;
   --colour-blue-80: #003170;
   --colour-blue-70: #004099;
```
chriddell, line 12: How come you only removed a select few colour vars?

creido-welly: I'm doing this over several PRs, a few at a time because there are a ****-tonne of 'em :)

chriddell: Ah okay, looks like the others are being removed in separate PRs?

Seems fine, might have been a bit easier to review this if all the color vars had been removed/replaced in this PR - that way it might have been clearer to see any issues - but it's not a big deal. 

I'll approve.

chriddell (approved):

PR #543, opened 2021-10-05
9031 task(style): Replaced local colour variables with design system theme variables

creido-welly:


chriddell (approved):
Same again... just checking you meant to open the PR against `task/9031-design-system-colours-3` (instead of `develop`)...

PR #544, opened 2021-10-06
8960 task: listing page accessibility

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8960

Manage focus on the listing pages:

- after user click on the pagination button, we want the page to scroll up and move the focus to the first result on the page.

Currently page scrolls up but the focus stays on the pagination.

### This PR

- adds forwardRef to components used on the listing pages


creido-welly (approved):
Phew, all good 😅 

Just for future, I find it helps the reviewers to point out which lines have been changed when file changes appear to be major but are really only as a result of increased nesting :)

skibinska:
thanks @creido-welly! that's a good point!

PR #556, opened 2021-11-11
9183 task: allows Label to accept markup

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/9183

### Context

Allow the `Label` component to accept the markup. We need this for a label with an asterix to be accessible for the screen reader users.

Usage
```html
<Label>
   {text}
   <span aria-hidden="true">*</span>
</Label>

or

<Label text={label} />
```

Screenshot 2021-11-11 at 15 14 24

### This PR
- allows the Label component to accept markup

### DIFF
--- a/src/components/Label/Label.tsx
+++ b/src/components/Label/Label.tsx
@@ -2,13 +2,15 @@ import React from 'react';
 import cx from 'classnames';
 
 type LabelProps = {
+  children?: React.ReactNode;
   className?: string;
   htmlFor: string;
   isDisabled?: boolean;
-  text: string;
+  text?: string;
 };
 
 export const FormLabel = ({
+  children,
   className,
   htmlFor,
   isDisabled,
@@ -21,7 +23,7 @@ export const FormLabel = ({
 
   return (
     <label className={classNames} htmlFor={htmlFor}>
-      {text}
+      {children || text}
     </label>
   );
 };
--- a/src/components/Label/Label.tsx
+++ b/src/components/Label/Label.tsx
@@ -2,15 +2,13 @@ import React from 'react';
 import cx from 'classnames';
 
 type LabelProps = {
-  children?: React.ReactNode;
   className?: string;
   htmlFor: string;
   isDisabled?: boolean;
-  text?: string;
+  text: string;
 };
 
 export const FormLabel = ({
-  children,
   className,
   htmlFor,
   isDisabled,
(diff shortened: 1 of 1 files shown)


src/components/Label/Label.tsx
```diff
@@ -21,7 +23,7 @@ export const FormLabel = ({
 
   return (
     <label className={classNames} htmlFor={htmlFor}>
-      {text}
+      {children || text}
```
creido-welly: For better security against XSS attacks it's prob worth doing this.
```
<label className={classNames} htmlFor={htmlFor} dangerouslySetInnerHTML=sanitizeHtml(children || text)} />
```

**Additional comments not for addressing immediately**
Just as an aside you could avoid introducing the children prop if you did this
```
<label className={classNames} htmlFor={htmlFor} dangerouslySetInnerHTML=sanitizeHtml(text)} />
```

... but then it would be much less readable in use when introducing markup i.e.
```
<Label text={`${variable}<span aria-hidden="true">*</span>`} />
```

Ideally we should replace `text` with `children` but I can see that's a lot of refactoring and doesn't need addressing in the scope of this PR. Maybe deprecate `text` and refactor in future?

skibinska: I tried the suggested approaches and I can't get the correct output:

1. 
```js
<label dangerouslySetInnerHTML={{ __html: sanitizeHtml(text) }} />
```
the output is not correct (I am missing aria-hidden), so the markup is
```html
<label class="">
  Label text
  <span>*</span>
</Label>
```
2. `sanitizeHtml` doesn't accept children (type `ReactNode`)
3. when I changed the type of children to string and passed
```html
 <Label className="c-newsletter__label" htmlFor={id}>
   {`${label}<span aria-hidden="true">*</span>`}
  </Label>
```
the aria-hidden still wasn't on the page.

skibinska: ~~I can do something like this for now:~~
Actually I can't do this as I am getting an error: 
> Error: Can only set one of `children` or `props.dangerouslySetInnerHTML`.

```html
<label
  dangerouslySetInnerHTML={{ __html: sanitizeHtml(text) }}
  className={classNames}
  htmlFor={htmlFor}
>
  {children}
</label>
```

creido-welly: Hmm. I suspect aria-hidden is getting stripped out by the `xss` plugin

I'm looking at options for extending it to allow aria-hidden

creido-welly: This PR should fix this problem 👍 

https://github.com/wellcometrust/corporate-components/pull/557

creido-welly (approved):

PR #563, opened 2021-12-02
9266 task: FormSelect accepts two types of options

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/9266

### This PR

- allows `FormSelect` to accept two types of options: an array of strings or an array of object (with `label` and `value`)

`FormSelect` with array of strings is used on the forms and with an array of objects on newsletter signup.

### DIFF
--- a/src/components/FormSelect/FormSelect.tsx
+++ b/src/components/FormSelect/FormSelect.tsx
@@ -1,13 +1,18 @@
 import React, { forwardRef } from 'react';
 import cx from 'classnames';
 
+type FormSelectOptionProps = {
+  label: string;
+  value?: string;
+};
+
 type FormSelectProps = {
   className?: string;
   defaultText?: string;
   id: string;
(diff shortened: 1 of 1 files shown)


src/components/FormSelect/FormSelect.tsx
```diff
@@ -38,9 +43,19 @@ export const FormSelect = forwardRef(
...
         </option>
-        {options.map(option => (
-          <option key={`option-${option}`}>{option}</option>
-        ))}
+        {options.map(option => {
+          if (typeof option === 'object') {
+            return (
+              <option
+                key={`option-${option.label}`}
+                value={option.value || option.label}
```
creido-welly: Are there likely to be any instances where there is a label but no value?

skibinska: probably not. I will update.

creido-welly (approved):

PR #564, opened 2021-12-07
9279 task: exports ExternalLinkMarker

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/9279

### Context

Some of the forms have internal links that we open in a new tab i.e. terms and conditions link on the [Returning from sick leave form](https://wellcome.org/grant-funding/guidance/sick-leave-form). We need a way of telling user that this link will be open in a new tab (icon and visually hidden text).

### This PR

- exports  `ExternalLinkMarker` so it can be used in CR - [example pR in CR](https://github.com/wellcometrust/corporate-react/pull/1184/files#diff-ebe697bf3d7a9ad193ec54a055f731730aab1a54bf676c6075ad37f0e3deb9d7R618)


src/index.ts
```diff
@@ -28,6 +28,7 @@ export { Image, ImageElement } from 'Image/Image';
 export { InfoBox } from 'InfoBox/InfoBox';
 export { default as Label } from 'Label';
 export { default as Link } from 'Link';
+export { ExternalLinkMarker } from 'Link';
```
skibinska: Is it a correct way, or should we do 
`export {Link, ExternalLinkMarker} from 'Link/Link'`

creido-welly: You can combine lines 30 and 31 as follows:

```
export { default as Link, ExternalLinkMarker } from 'Link';
```

skibinska: Updated :)

creido-welly (approved):

PR #566, opened 2021-12-09
8905 bug: Prevent gaps caused by long image captions

creido-welly:
Relates to wellcometrust/corporate#8905

Allows right aligned image (and video) caption to overlap the content row beneath to prevent excess gapping due to CSS grid (screen width of 1024px and above). The caveat to this work is that there must be sufficient content to prevent overlap of other content which occupies the right hand column. The content team have been made aware of this.

Current layout showing CSS grid:
Screenshot 2021-12-09 at 15 15 20

Modified layout:
Screenshot 2021-12-09 at 15 14 30

### DIFF
--- a/src/components/Media/_media.scss
+++ b/src/components/Media/_media.scss
@@ -29,21 +29,14 @@
 
 .cc-media__caption {
   align-self: start;
-  // D7 override
-  border: 0;
-  border-left: 1px solid var(--color-grey-70);
-  color: var(--color-grey-70);
-  font-size: var(--body-xxs);
-  line-height: var(--body-line-height);
   margin-top: calc(2 * var(--space-unit));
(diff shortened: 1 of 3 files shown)


src/components/Media/_media.scss
```diff
@@ -29,21 +29,14 @@
...
   margin-top: calc(2 * var(--space-unit));
-  padding: calc(0.25 * var(--space-unit)) 0 calc(0.25 * var(--space-unit)) calc(2 * var(--space-unit));
 
   a {
     @include animated-link;
   }
 }
 
-.cc-media__caption--below-image {
+.cc-media--caption--below-image {
```
skibinska: `cc-media--caption-below-image` (we should remove an extra hyphen).

creido-welly: Doh!

creido-welly: Fixed

skibinska (approved):

PR #570, opened 2022-01-05
8904 bug: fixes gap in the header when the sidebar longer than content

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/8904

### Context

The two authors in the header increase the header size and the same happens with the image caption when the screen width narrows (1024px up to about 1340px).

Screenshot 2022-01-05 at 16 13 50

###  This PR

- fixes this issue

Screenshot 2022-01-05 at 16 15 23

### DIFF
--- a/src/components/PageHeaderCompact/_page-header-compact.scss
+++ b/src/components/PageHeaderCompact/_page-header-compact.scss
@@ -18,6 +18,7 @@
     display: grid;
     // stylelint-disable value-no-vendor-prefix
     display: -ms-grid;
+    position: relative;
   }
 
   @include mq(xl) {
@@ -72,6 +73,7 @@
     margin-top: calc(var(--body-sm) +
     (2 * var(--space-unit)) +
     (var(--body-line-height) * var(--body-sm) * 0.5));
+    position: absolute;
   }
 }
 


src/components/PageHeaderCompact/_page-header-compact.scss
```diff
@@ -72,6 +73,7 @@
     margin-top: calc(var(--body-sm) +
     (2 * var(--space-unit)) +
     (var(--body-line-height) * var(--body-sm) * 0.5));
+    position: absolute;
```
creido-welly, line 76: What can I say, it works and it's not the first time we've fixed grid gaps using `position: absolute`. I believe all of the following are true so for the time-being it's prob fine (but you should also test this as you mentioned):
* authors and topics are only available in PageHeaderCompact
* PageHeaderCompact is only used on article page types
* InpageNav is not used on article page types

All that said, I have real reservations about using this approach because it is effectively a hack and if it ever came to a situation where we want e.g. authors, topics and in-page nav there is a risk of overlap. Long-term we need to look at how we can refactor page headers, perhaps so that sidebar components are all at the same level and therefore their vertical positions are determined by each other. I also suggest the dev team prototype this for the designers to illustrate the problems we currently have. I'll create a separate ticket.

creido-welly (approved):
Approved subject to verifying presence of in-page nav on article pages

skibinska:
If in-page nav is on the article page it overlaps on the topics/authors sidebar. 

I will close this ticket, as I don't want to introduce any bugs.

Screenshot 2022-01-06 at 15 49 53

creido-welly:
@skibinska new discovery ticket created for dealing with this issue - https://github.com/wellcometrust/corporate/issues/9325

PR #574, opened 2022-01-07
9329 bug(style): fixes style within Info box

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/9329

### This PR

- removes InfoBox related classes
- removes Modal component


src/components/Accordion/_accordion.scss
```diff
@@ -98,7 +98,7 @@
     padding-right: var(--space-lg);
   }
 
-  .cc-info-box & {
+  .c-info-box & {
```
skibinska: @creido-welly, do you think it would be better to move that (and below) style to corporate-react?

src/components/Modal/Modal.tsx
```diff
@@ -34,7 +34,7 @@ export const Modal = ({
 }: ModalProps) => {
   const [wasAccepted, setWasAccepted] = useState(false);
 
-  const classNames = cx('cc-info-box cc-modal', {
+  const classNames = cx('c-info-box cc-modal', {
```
skibinska: @creido-welly, should I just copy-pase the infobox styles and add them to `cc-modal`?

src/components/Accordion/_accordion.scss
creido-welly: Yes move this style to infobox stylesheet in corporate-react

src/components/Button/_button.scss
```diff
@@ -119,7 +119,7 @@
   text-align: center;
 }
 
-.cc-info-box .cc-cta {
+.c-info-box .cc-cta {
```
creido-welly: This is all very confusing isn't it. `cc-cta` is only applied to JSX in corporate-react so I suggest moving everything below line 116 in this file to corporate-react. It should really be refactored as `.c-cta` too.

src/components/Modal/Modal.tsx
creido-welly: No I don't think you need to do that. On the other hand Modal needs moving to corporate-react at some point anyway. If it's not too big a job to handle right now it could be done in this PR. Modal component doesn't have any dependents in this repo FYI.

https://www.notion.so/wellcometrust/wellcome-org-frontend-component-audit-28be0bdf2b4e4d8580cc1f6afac64da1

creido-welly (commented):
Comments left re style rule locations

creido-welly (approved):

PR #584, opened 2022-02-03
9156 task: fixed non-descriptive person links

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/9156

### This PR

- adds aria-label to make person social links more descriptive to screen reader

### DIFF
--- a/src/utils/get-link-label.ts
+++ b/src/utils/get-link-label.ts
@@ -1,17 +1,29 @@
-export const getLinkLabel = (link: string) => {
-  const findMatch = link.match(
+/**
+ * Create an accessible link text for an author's social links and contact details.
+ *
+ * Example:
+ * <a aria-label={getLinkLabel(name, link)} href='twitter.com/author'>Twitter</a>
+ *
+ * @param name
+ * @param link
+ */
+export const getLinkLabel = (
+  name: string,
+  link: { url: string; title: string }
+) => {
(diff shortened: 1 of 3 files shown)


src/utils/get-link-label.ts
```diff
@@ -0,0 +1,29 @@
...
+ * <a aria-label={getLinkLabel(name, link)} href='twitter.com/author'>Twitter</a>
+ *
+ * @param name
+ * @param link
+ */
+export const getLinkLabel = (
+  name: string,
+  link: { url: string; title: string }
+) => {
+  const findMatch = link.url.match(
```
creido-welly: This list is unlikely to change but it's worth mentioning that testing against a regex which is separate to linkMap's keys needs both parts updating. It's always better to work from a single source to avoid bugs.

One way of doing this (and this is similar to the funding scheme listing page) would be to use `Object.keys` and then `Array.find` to return a matching value

https://codepen.io/creido-welly/pen/JjORWoG?editors=0011

```
Object.keys(linkMap).find(key => str.includes(key))
```

As an aside, I think you should also include the colons for mailto: and tel: in the keys (I've done this in the codepen)

skibinska: Good point. I will update.

creido-welly (approved):

PR #586, opened 2022-02-08
9466 task: adds ReadTime component

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/9466

### Context

Add time to read to an article and to article's cards.

### This PR

- adds `ReadTime` component
- updates card components and PageHeaderCompact to use ReadTime

Related CR PR: https://github.com/wellcometrust/corporate-react/pull/1242

_Example_

### DIFF
--- a/src/components/ReadTime/ReadTime.tsx
+++ b/src/components/ReadTime/ReadTime.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import cx from 'classnames';
+import FormattedDate from 'FormattedDate';
+
+type ReadTimeProps = {
+  className?: string;
+  date?: string;
+  metaLabel?: string;
+  readTime?: string;
+};
+
+export const ReadTime = ({
+  className,
+  date,
+  metaLabel,
+  readTime
+}: ReadTimeProps) => {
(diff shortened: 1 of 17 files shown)


src/components/ReadTime/ReadTime.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const ReadTime = ({
+  className,
+  date,
+  metaLabel,
+  readTime
+}: ReadTimeProps) => {
+  return (
+    <div className={cx('cc-read-time', { [className]: className })}>
```
creido-welly, line 19: It's perhaps an edge case that readTime is falsy but it might be worth adding logic for handling that in the render.

creido-welly (commented):
Code is fine as far as I can see. One logic handling issue to address.

I can see why you have added this component to corporate-components since there are so many related component types. but in the long-term I think it would have been less development work to add this to corporate-react.

We're adding ReadTime in one giant hit here which is always more risky. Better to progressively add ReadTime to each component in turn as it gets moved across to CR and then this can be done in smaller chunks.

PR #601, opened 2022-02-17
9539 bug: fix to prevent anchor id being removed by xss filter

creido-welly:
Relates to wellcometrust/corporate#9539


skibinska:
It wasn't a good idea to move `sanitizeHTML` to CR, as now we have two places to maintain. I am going to remove it from CR.

skibinska (approved):

PR #610, opened 2022-02-25
9583 task: adds audio player icons

skibinska:
Relates https://github.com/wellcometrust/corporate/issues/9583

### This PR

- adds audio player icons: play, pause, volume and mute.

### DIFF
--- a/src/components/Icon/components/Mute.tsx
+++ b/src/components/Icon/components/Mute.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+
+import { IconSVGProps } from '../Icon';
+
+const Mute = (props: IconSVGProps) => (
+  <svg viewBox="0 0 24 24" {...props} xmlns="http://www.w3.org/2000/svg">
+    <g fill="none" fillRule="evenodd">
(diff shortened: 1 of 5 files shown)


creido-welly:
I think these icons need to be added to the [Design System](https://designsystem.wellcome.org/?path=/story/global-icons--page) and then imported into CR from there. The CC icons will be swapped out for DS ones at some point (ticket wellcometrust/corporate#9034) but that can happen separately.

## DS
When adding the new icons follow the naming convention already established - I would add them as `iconAction...`.

## CR
You will need to npm install the DS package, import the component into React and also import the css in [60-components/__manifest.scss](https://github.com/wellcometrust/corporate-react/blob/develop/src/styles/60-components/__manifest.scss)

```
@import '~@wellcometrust/design-system/dist/components/Icon/Icon';
```

creido-welly:
As discussed just now, this work can be parked until we know more about the approach we will use for "listen to article"

skibinska:
PR in design system created: https://github.com/wellcometrust/design-system/pull/32

